PR #1, opened 2022-05-02
Add License file

derekparker:
Adds initial LICENSE file.

Wanted to have this as it's own commit so we can agree on Licencing, copyright statement, etc.


dbenoit17 (approved):
lgtm

PR #2, opened 2022-05-03
Add initial code from golang-fips/go

derekparker:
This PR adds initial code from the golang-fips/go repo. It takes the latest iteration of our changes and puts them in their own package, `openssl`, which we will later import into the main Go crypto packages via a patch to be carried.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,3 @@
+module github.com/openssl-fips
+
+go 1.18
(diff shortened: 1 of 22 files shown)


dbenoit17 (approved):
LGTM, tested against OpenSSL 1.1 and OpenSSL 3.

PR #4, opened 2022-05-05
openssl: Add Unreachable and UnreachableExceptTest

derekparker:
Adding these functions back in because they will be used by Go once we
patch this package in in place of the boring package.


derekparker:
Closing this out as it won't actually be necessary.

PR #14, opened 2022-09-12
Add Go 1.19 support

derekparker:
There are more changes necessary to support Go 1.19 and many of them are backwards incompatible as there are function signature changes. This patch introduces a series of changes to support Go 1.19 but will not be backwards compatible with older versions.

This is acceptable because we only intend to support Go 1.19+ using this particular architecture.


ueno (approved):

dbenoit17 (approved):
tested, lgtm

PR #15, opened 2022-09-20
Implement OpenSSL loading and RandReader

qmuntal:
This PR contains three major features: portable OpenSSL, dynamic OpenSSL version selection, and random number generator.

I've also added some minimum GH Actions infra for testing that everything works using the platform OpenSSL.

### Portable and header-less OpenSSL loading

The OpenSSL loading differs between Microsoft and RedHat, even though both define the OpenSSL C functions via `DEFINEFUNC`-style macros.

The fundamental difference is that RH uses `OPENSSL_VERSION_NUMBER` (from the OpenSSL headers) to decide which version is being loaded, but MSFT calls an OpenSSL function to get it at runtime. The benefit of the MSFT approach is that it does not depend on which header was included at compile time; one could build an app using OpenSSL 1.1.1 and then run it on a different environment with OpenSSL 3.0.0. On the other hand, `OPENSSL_VERSION_NUMBER` would always report the same version number, the one used to build the app, so this scenario wouldn't work.

The same happens with other OpenSSL header `#defines`, we can't use them if we want to support OpenSSL portability. Therefore, the idea is that we don't include any OpenSSL header but define everything on our headers. On a follow-up PR I'll submit a tool that checks that our definitions match with OpenSSL headers (this tool exists in MSFT and RH repos, I'll see how it can be merged). Notice that Google follows this same approach for their BoringSSL bindings.

### Dynamic OpenSSL version selection

The OpenSSL shared library name does not follow a well-defined standard. For example, some platforms name OpenSSL 1.1.1 `libcrypto.so.1.1` and others `libcrypto.so.111`.

In order to be as generic as possible, this PR implements an OpenSSL version selection algorithm inherited from .Net Runtime, which also has to support many different platforms:

- `openssl.Init(version)` must be called before any other OpenSSL call. There is no implicit initialization.

- If `version` is not empty, its value will be appended to the OpenSSL shared library name as a version suffix when calling `dlopen`. For example, `version=1.1.1k-fips` makes `Init` look for the shared library `libcrypto.so.1.1.1k-fips`.

- If `version` is empty, `Init` will try to load the OpenSSL shared library using a list of supported and well-known version suffixes, going from higher to lower versions. This version list includes the RedHat naming conventions. 

### Random number generator via `openssl.RandReader`

Both Microsoft and RedHat bindings had almost the same implementation, so this is a clean port. The only difference is that RH enhances the error more context by using `NewOpenSSLError`, function that I still haven't implemented in this branch. IMO it makes sense to give a detailed error description in such an important function, I'll do so in a follow-up PR.


openssl/init.go
```diff
@@ -0,0 +1,103 @@
...
+// FreeBSD library version numbering does not directly align to the version of OpenSSL.
+// Its preferred search order is 11 -> 111.
+//
+// Some distributions use 1.0.0 and others (such as Debian) 1.0.2 to refer to the same OpenSSL 1.0.2 version.
+//
+// Fedora derived distros use different naming for the version 1.0.x.
+var knownVersions = [...]string{"3", "1.1", "11", "111", "1.0.2", "1.0.0", "10"}
+
+// opensslInit loads and initialize OpenSSL.
+// If succeed, it returns the major and minor OpenSSL version
```
derekparker, line 25: s/succeed/successful/

openssl/init.go
```diff
@@ -0,0 +1,103 @@
...
+// as reported by the OpenSSL API.
+//
+// See Init() for details about version.
+func opensslInit(version string) (major int, minor int, err error) {
+	// Load the OpenSSL shareed library using dlopen.
```
derekparker, line 30: s/shareed/shared/

derekparker (changes requested):
LGTM, just a few spelling nits.

derekparker (approved):
LGTM

openssl/openssl.go
```diff
@@ -0,0 +1,56 @@
...
+// If version is empty, Init will try to load the OpenSSL shared library
+// using a list of supported and well-known version suffixes, going from higher to lower versions.
+func Init(version string) error {
+	initOnce.Do(func() {
+		vMajor, vMinor, initErr = opensslInit(version)
+	})
+	return initErr
+}
+
+func errUnsuportedVersion() error {
```
dagood, line 45:
```suggestion
func errUnsupportedVersion() error {
```

openssl/thread_setup.c
```diff
@@ -0,0 +1,46 @@
...
+
+#include <stdio.h>
+#include <pthread.h>
+#include <sys/types.h>
+#include <sys/syscall.h>
+
+#define _GNU_SOURCE
+#include <unistd.h>
+ 
+#define CRYPTO_LOCK      0x01
```
dagood, line 14: This used to be aligned with some adjacent lines, but they aren't here anymore:

```suggestion
#define CRYPTO_LOCK 0x01
```

openssl/thread_setup.c
```diff
@@ -0,0 +1,46 @@
...
+int go_openssl_thread_setup(void)
+{
+    mutex_buf = malloc(go_openssl_CRYPTO_num_locks()*sizeof(pthread_mutex_t));
+    if(!mutex_buf) {
+        return 0;
+    }
+    int i;
+    for(i = 0;  i < go_openssl_CRYPTO_num_locks(); i++) {
+        pthread_mutex_init(&mutex_buf[i], NULL);
+    }
```
dagood, line 42: A few changes to match the style in `openssl/goopenssl.c`:

```suggestion
    for (i = 0; i < go_openssl_CRYPTO_num_locks(); i++)
        pthread_mutex_init(&mutex_buf[i], NULL);
```

(And similar for the other blocks in the file.)

I personally prefer the style of always including `{` and `}` even for one statement, but one way or another, we should probably pick one and stick to it?

openssl/goopenssl.h
```diff
@@ -0,0 +1,34 @@
...
+int go_openssl_thread_setup(void);
+void go_openssl_load_functions(void* handle, int major, int minor);
+
+// Define pointers to all the used OpenSSL functions.
+// Calling C function pointers from Go is currently not supported.
+// It is possible to circumvent this by using a C function wrapper.
+// https://pkg.go.dev/cmd/cgo
+#define DEFINEFUNC(ret, func, args, argscall)      \
+    extern ret (*_g_##func)args;                   \
+    static inline ret go_openssl_##func args       \
```
ueno, line 18: If possible (depending on the C compilers we support), I would suggest using the `typeof` keyword so we don't need to spell out function signatures. That would also prevent any mismatches in function signatures between `goopenssl.h` and OpenSSL headers, as they should be caught at compile time.

[Here](https://github.com/ueno/go/commit/b85fb8abe0298b556a209f10de99eeb0a1239682) is my previous attempt; another example can be found in [libbpf](https://github.com/libbpf/libbpf/blob/10a32130e727719a69e1b2b10815af7c2e245be7/src/usdt.bpf.h#L235).

openssl/openssl.go
ueno, line 45: `errUnsuportedVersion` → `errUnsupportedVersion` ?

openssl/goopenssl.h
qmuntal: A couple of thoughts:
- CGO internally uses `__typeof__` to generate C bindings, so it can't be used with a C compiler which doesn't support that operand. That's the case of MSVC.
- I like the `typeof` approach, but this PR defines the OpenSSL symbols on a way that OpenSSL are not needed to facilitate OpenSSL portability (e.g., building with 1.1.1 and running with 3.0.1), so I'm not sure if `typeof` would work here.

openssl/thread_setup.c
```diff
@@ -0,0 +1,43 @@
...
+    return (unsigned long)syscall(__NR_gettid);
+}
+ 
+int go_openssl_thread_setup(void)
+{
+    mutex_buf = malloc(go_openssl_CRYPTO_num_locks()*sizeof(pthread_mutex_t));
+    if(!mutex_buf)
+        return 0;
+    int i;
+    for(i = 0;  i < go_openssl_CRYPTO_num_locks(); i++)
```
dagood, line 38: I didn't really mention that I also changed some spacing here to match style, so I think that bit got missed. (Adding a suggestion for each one I saw.)

```suggestion
    for (i = 0; i < go_openssl_CRYPTO_num_locks(); i++)
```

openssl/thread_setup.c
```diff
@@ -0,0 +1,43 @@
...
+ 
+static unsigned long id_function(void)
+{
+    return (unsigned long)syscall(__NR_gettid);
+}
+ 
+int go_openssl_thread_setup(void)
+{
+    mutex_buf = malloc(go_openssl_CRYPTO_num_locks()*sizeof(pthread_mutex_t));
+    if(!mutex_buf)
```
dagood, line 35:
```suggestion
    if (!mutex_buf)
```

openssl/thread_setup.c
```diff
@@ -0,0 +1,43 @@
...
+#include <unistd.h>
+ 
+#define CRYPTO_LOCK 0x01
+
+/* This array will store all of the mutexes available to OpenSSL. */ 
+static pthread_mutex_t *mutex_buf = NULL;
+ 
+static void locking_function(int mode, int n, const char *file, int line)
+{
+    if(mode & CRYPTO_LOCK)
```
dagood, line 21:
```suggestion
    if (mode & CRYPTO_LOCK)
```

dagood (approved):
LGTM, just a few whitespace style suggestions. 😄

PR #19, opened 2022-09-27
Implement FIPS() and SetFIPS()

qmuntal:
This PR adds two new exported functions, `FIPS() bool` and `SetFIPS(enabled bool) error`, both ported from `microsoft/go-crypto-openssl`.

### func FIPS

`golang-fips/openssl-fips` also implements `FIPS()` in the unexported [fipsModeEnabled](https://github.com/golang-fips/openssl-fips/blob/60f04d7f65e2dacaaed870f38db900807caf37f0/openssl/openssl.go#L82). The problem is that ir relies on `FIPS_mode()`, which works on OpenSSL 1.x,  but has been removed from OpenSSL 3. RedHat could probably use it because their OpenSSL 3 binaries are patched to support this function, but we can't generally assume it will be there.

I've used instead a combination of tips explained in https://wiki.openssl.org/index.php/OpenSSL_3.0#Using_the_FIPS_Module_in_applications:
- First check [EVP_default_properties_is_fips_enabled()](https://www.openssl.org/docs/man3.0/man3/EVP_default_properties_enable_fips.html) == true
- Then check that `EVP_MD_fetch(NULL, "SHA2-256", "fips=yes")` succeed.

### func FIPS

On OpenSSL 1.x uses the good-old `FIPS_mode_set(1)`. On OpenSSL 3 it tries to load the FIPS provider, if not already loaded, and updates the default EVP properties to use FIPS.

### DIFF
--- a/openssl/goopenssl.c
+++ b/openssl/goopenssl.c
@@ -11,14 +11,18 @@
 
 #define DEFINEFUNC(ret, func, args, argscall)                  ret (*_g_##func)args;
 #define DEFINEFUNC_LEGACY_1_0(ret, func, args, argscall)       DEFINEFUNC(ret, func, args, argscall)
+#define DEFINEFUNC_LEGACY_1(ret, func, args, argscall)         DEFINEFUNC(ret, func, args, argscall)
 #define DEFINEFUNC_1_1(ret, func, args, argscall)              DEFINEFUNC(ret, func, args, argscall)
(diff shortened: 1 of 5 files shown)


dagood (approved):

derekparker (approved):
LGTM but I'd also like to get @ueno +1 on it.

ueno (approved):
Looks good to me.

PR #20, opened 2022-09-27
Lower version requirement in go.mod to Go 1.16

dbenoit17:
Users should be able to use older toolchain versions to bootstrap newer distributions. This change lowers the version requirement for the host compiler to Go 1.16.


derekparker:
I'm not sure about this. In fact I think we should increase it to `go1.19` since the code in this repo, as it stands, is incompatible with any older version of Go.

PR #28, opened 2022-10-05
Implement newOpenSSLError

qmuntal:
This PR implements `newOpenSSLError` using RedHat code with the following modifications:

- Do not export the function, there is no need to call it from outside.
- Use plain string concatenation and `strconv` instead of `fmt` to construct the error message. Upstream try to avoid importing `fmt` in low level crypto packages, so do we.
-  RedHat code appends the error flags to the error string. These flags don't give information about the error, but about the error data, which we don't use. Therefore, we can use `ERR_get_error_line` instead of `ERR_get_error_line_data` and skip also the flags.
-  The formatted error string has been modified to be more *idiomatic*. Example:
```
panic: FIPS_mode_set
openssl error(s):
error:0F06D065:common libcrypto routines:FIPS_mode_set:fips mode not supported
        ../crypto/o_fips.c:22
```

instead of 

```
panic: FIPS_mode_set
openssl error(s):
file: ../crypto/o_fips.c
line: 22
function: unknown
flags: 0
error string: error:0F06D065:common libcrypto routines:FIPS_mode_set:fips mode not supported
```

### DIFF
--- a/openssl/openssl.go
+++ b/openssl/openssl.go
@@ -10,6 +10,7 @@ import "C"
 import (
 	"errors"
 	"strconv"
+	"strings"
 	"sync"
 	"unsafe"
 )
@@ -114,7 +115,7 @@ func SetFIPS(enabled bool) error {
 			mode = C.int(0)
 		}
 		if C.go_openssl_FIPS_mode_set(mode) != 1 {
-			return fail("FIPS_mode_set")
+			return newOpenSSLError("FIPS_mode_set")
 		}
 		return nil
 	case 3:
@@ -130,7 +131,7 @@ func SetFIPS(enabled bool) error {
 		if !providerAvailable(props) {
 			// If not, fallback to provName provider.
 			if C.go_openssl_OSSL_PROVIDER_load(nil, provName) == nil {
-				return fail("OSSL_PROVIDER_try_load")
+				return newOpenSSLError("OSSL_PROVIDER_try_load")
 			}
 			// Make sure we now have a provider available.
 			if !providerAvailable(props) {
@@ -138,7 +139,7 @@ func SetFIPS(enabled bool) error {
 			}
 		}
 		if C.go_openssl_EVP_set_default_properties(nil, props) != 1 {
-			return fail("EVP_set_default_properties")
+			return newOpenSSLError("EVP_set_default_properties")
 		}
 		return nil
 	default:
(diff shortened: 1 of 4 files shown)


openssl/rand.go
```diff
@@ -14,7 +14,7 @@ func (randReader) Read(b []byte) (int, error) {
 	// We check it even so.
 	if len(b) > 0 && C.go_openssl_RAND_bytes((*C.uchar)(unsafe.Pointer(&b[0])), C.int(len(b))) == 0 {
 		// TODO: use NewOpenSSLError once implemented.
```
dagood, line 16:
```suggestion
```

openssl/openssl.go
```diff
@@ -129,18 +131,47 @@ func SetFIPS(enabled bool) error {
...
+		case 3:
+			e = C.go_openssl_ERR_get_error_all(&file, &line, nil, nil, nil)
+		default:
+			panic(errUnsupportedVersion())
+		}
+		if e == 0 {
+			break
+		}
+		b.WriteByte('\n')
+		var buf [256]byte
```
dagood, line 172: Is 256 a documented max (or recommendation) or is it just good enough? (This may go back to the original code.)

openssl/openssl.go
```diff
@@ -129,18 +131,47 @@ func SetFIPS(enabled bool) error {
...
+		C.go_openssl_ERR_error_string_n(e, (*C.char)(unsafe.Pointer(&buf[0])), C.size_t(len(buf)))
+		b.WriteString(string(buf[:]) + "\n\t" + C.GoString(file) + ":" + strconv.Itoa(int(line)))
```
dagood, line 174: Since we're already using a builder, would it be more efficient to add these pieces one-by-one with `b.WriteString`?

openssl/openssl.go
qmuntal: It is a documented recommendation which should be good enough: https://www.openssl.org/docs/manmaster/man3/ERR_error_string_n.html.

openssl/openssl.go
qmuntal: It would be more efficient, but also uglier:

```go
b.WriteString(string(buf[:]) + "\n\t" + C.GoString(file) + ":" + strconv.Itoa(int(line)))

// vs

b.WriteByte('\n')
b.Write(buf[:])
b.WriteString("\n\t")
b.WriteString(C.GoString(file))
b.WriteByte(':')
b.WriteString(strconv.Itoa(int(line)))
```

I would rather keep it one-liner 😄

dagood (approved):

derekparker (approved):
LGTM

PR #30, opened 2022-10-07
Implement AES ciphers

qmuntal:
⚠️ Before submitting this PR i rogue-pushed `aes.go` from master into `wip/combined-codebase-staging` so the changes between RedHat and Microsoft are easier to see.

This PR is almost a direct port from Microsoft repo, which was heavily based on RedHat implementation. I've formatted a little bit the Microsoft code to minimize differences with RedHat.

I'll comment inline the changes between both codebases, there are a few lines that I'm not sure which version to use.

There is one big addition coming from Microsoft repo: `newGCM(tls bool)` honors the TLS mode by enforcing strictly monotonically increasing nonces and also fail after reusing the same key 2^64 - 1 keys times. See [this document](https://github.com/microsoft/go-crypto-openssl/blob/main/docs/go-openssl-compat.md#aes-gcm-encryption-in-tls-mode) to understand how we get there. Feel free to push back on any detail.


openssl/aes.go
```diff
@@ -116,32 +107,19 @@ func (c *aesCipher) Decrypt(dst, src []byte) {
...
-
-		if C.int(1) != C._goboringcrypto_EVP_CipherInit_ex(c.dec_ctx, c.cipher, nil, k, nil, C.GO_AES_DECRYPT) {
-			panic("cipher: unable to initialize EVP cipher ctx")
+		var err error
+		c.dec_ctx, err = newCipherCtx(c.cipher, C.GO_AES_DECRYPT, c.key, nil)
+		if err != nil {
+			panic(err)
 		}
 	}
-	// Workaround - padding detection is broken but we don't need it
```
qmuntal, line 130: I haven't found the need to implement this workaround. Could you give more context or share a failing test case?

openssl/aes.go
```diff
@@ -157,17 +135,7 @@ func (x *aesCBC) CryptBlocks(dst, src []byte) {
 		panic("crypto/cipher: output smaller than input")
 	}
 	if len(src) > 0 {
-		outlen := C.int(0)
-		// Workaround - padding detection is broken but we don't need it
-		// since we check for full blocks
-		if C._goboringcrypto_EVP_CIPHER_CTX_set_padding(x.ctx, 0) != 1 {
```
qmuntal, line 163: @ueno @dbenoit17 I haven't found the need to implement this workaround. Could you give more context or share a failing test case?

openssl/aes.go
```diff
@@ -116,32 +107,19 @@ func (c *aesCipher) Decrypt(dst, src []byte) {
...
 	}
-	// Workaround - padding detection is broken but we don't need it
-	// since we check for full blocks
-	if C._goboringcrypto_EVP_CIPHER_CTX_set_padding(c.dec_ctx, 0) != 1 {
-		panic("crypto/cipher: could not disable cipher padding")
-	}
-	outlen := C.int(0)
-	C._goboringcrypto_EVP_CipherUpdate(c.dec_ctx, (*C.uchar)(unsafe.Pointer(&dst[0])), &outlen, (*C.uchar)(unsafe.Pointer(&src[0])), C.int(aesBlockSize))
+
+	C.go_openssl_EVP_DecryptUpdate_wrapper(c.dec_ctx, base(dst), base(src), aesBlockSize)
```
qmuntal, line 117: FYI: all `C.go_openssl_EVP_*_wrapper` functions are there just to avoid allocating `outlen` just to discard it without being used.

openssl/aes.go
```diff
@@ -178,37 +146,29 @@ func (x *aesCBC) SetIV(iv []byte) {
...
-	copy(x.iv[:], iv)
-	if C.int(1) != C._goboringcrypto_EVP_CipherInit_ex(x.ctx, nil, nil, nil, (*C.uchar)(unsafe.Pointer(&x.iv[0])), -1) {
+	if C.go_openssl_EVP_CipherInit_ex(x.ctx, nil, nil, nil, base(iv), -1) != 1 {
 		panic("cipher: unable to initialize EVP cipher ctx")
 	}
 }
 
 func (c *aesCipher) NewCBCEncrypter(iv []byte) cipher.BlockMode {
-	x := &aesCBC{key: c.key, mode: C.GO_AES_ENCRYPT}
-	copy(x.iv[:], iv)
```
qmuntal, line 189: I'm not copying `iv` here because `EVP_CipherInit_ex` already copies it before storing.

openssl/aes.go
```diff
@@ -272,39 +222,28 @@ func (x *aesCTR) XORKeyStream(dst, src []byte) {
...
-		(*C.uint8_t)(unsafe.Pointer(&src[0])),
-		(*C.uint8_t)(unsafe.Pointer(&dst[0])),
-		C.size_t(len(src)))
+	C.go_openssl_EVP_EncryptUpdate_wrapper(x.ctx, base(dst), base(src), C.int(len(src)))
 	runtime.KeepAlive(x)
 }
 
 func (c *aesCipher) NewCTR(iv []byte) cipher.Stream {
-	x := &aesCTR{key: c.key}
-	copy(x.iv[:], iv)
```
qmuntal, line 285: I'm not copying `iv` here because `EVP_CipherInit_ex` already copies it before storing.

openssl/aes.go
```diff
@@ -338,27 +279,15 @@ type noGCM struct {
 }
 
 func (c *aesCipher) NewGCM(nonceSize, tagSize int) (cipher.AEAD, error) {
-	if !ExecutingTest() || IsStrictFips() {
```
qmuntal, line 341: `IsStrictFips` should be moved outside this library and implemented as part of its integration with Go (i.e, in the crypto backend).

openssl/aes.go
```diff
@@ -374,28 +303,30 @@ func (c *aesCipher) NewGCMTLS() (cipher.AEAD, error) {
 }
 
 func (c *aesCipher) newGCM(tls bool) (cipher.AEAD, error) {
-	keyLen := len(c.key) * 8
-
-	if keyLen != 128 && keyLen != 256 {
-		if ExecutingTest() {
-			// Fall back to standard library for GCM with non-standard key size.
```
qmuntal, line 381: This PR implements AES 128 GCM cipher, so all upstream tests will be happy.

openssl/aes.go
```diff
@@ -374,28 +303,30 @@ func (c *aesCipher) NewGCMTLS() (cipher.AEAD, error) {
...
-	}
-
-	g := &aesGCM{key: c.key, tls: tls}
-	if g.NonceSize() != gcmStandardNonceSize {
-		panic("boringcrypto: internal confusion about nonce size")
+	var cipher C.GO_EVP_CIPHER_PTR
+	switch len(c.key) * 8 {
+	case 128:
+		cipher = C.go_openssl_EVP_aes_128_gcm()
+	case 192:
```
qmuntal, line 310: 192-bit keys are not supported by RedHat. This restirction should me moved to the integration with Go (i.e, in the crypto backend).

openssl/aes.go
```diff
@@ -423,34 +345,49 @@ func (g *aesGCM) Seal(dst, nonce, plaintext, additionalData []byte) []byte {
...
-		(*C.uint8_t)(unsafe.Pointer(&dst[n])),
-		base(nonce), base(additionalData), C.size_t(len(additionalData)),
-		base(plaintext), C.size_t(len(plaintext)), &ciphertextLen,
-		base(g.key), C.int(len(g.key)*8)); ok != 1 {
-		panic("boringcrypto: EVP_CIPHER_CTX_seal fail")
+		panic(fail("EVP_CIPHER_CTX_seal"))
 	}
 	runtime.KeepAlive(g)
-
-	if ciphertextLen != C.size_t(len(plaintext)+gcmTagSize) {
```
qmuntal, line 450: This check is done inside `go_openssl_EVP_CIPHER_CTX_seal_wrapper`.

openssl/aes.go
```diff
@@ -465,43 +402,60 @@ func (g *aesGCM) Open(dst, nonce, ciphertext, additionalData []byte) ([]byte, er
...
 	if ok == 0 {
 		// Zero output buffer on error.
-		for i := range dst {
-			dst[i] = 0
+		for i := range out {
+			out[i] = 0
 		}
 		return nil, errOpen
 	}
-	if outLen != C.size_t(len(ciphertext)-gcmTagSize) {
```
qmuntal, line 499: This check is done inside `go_openssl_EVP_CIPHER_CTX_open_wrapper`.

openssl/aes.go
derekparker: The original catalyst for this change if I'm not mistaken is due to: https://bugzilla.redhat.com/show_bug.cgi?id=1979100. A lot of the discussion on that thread is private but that is what led to the change.

qmuntal: Thanks for the context. I can't reproduce the issue with the current implementation, but I'm fine adding the workaround just in case. Anyway, it makes sense to explicitly disable padding if Go expects full blocks.

Just one comment: is there any reason `EVP_CIPHER_CTX_set_padding` is not inside the previous block, which is just called once to initialize the context?

openssl/aes.go
qmuntal: @derekparker could we move this `EVP_CIPHER_CTX_set_padding` to `NewCBCDecrypter` and `NewCBCEncrypter` so we only call it once per `aesCBC` instance instead of once per `CryptBlock` call?

dagood (approved):
I also compared from microsoft/go-crypto-openssl -> this branch so I could also look at it from the perspective of what has changed from what we've been using in microsoft/go. Didn't see anything unexpected. LGTM!

openssl/aes.go
derekparker: I don't believe so, it should be able to be hoisted.

openssl/aes.go
derekparker: I think that should be fine.

derekparker:
Sorry for the delay in reviewing. I will take a full review pass once the above comments are sorted out (moving around certain openssl calls).

openssl/aes.go
qmuntal: Done

openssl/aes.go
qmuntal: Done.

openssl/aes.go
derekparker: Agreed.

derekparker (approved):
LGTM

qmuntal:
@derekparker should we wait for @ueno?

derekparker:
@qmuntal I would prefer to. I will ping him for a review on this.

openssl/aes.go
```diff
@@ -88,20 +85,14 @@ func (c *aesCipher) Encrypt(dst, src []byte) {
...
+		var err error
+		c.enc_ctx, err = newCipherCtx(c.cipher, C.GO_AES_ENCRYPT, c.key, nil)
+		if err != nil {
+			panic(err)
 		}
 	}
 
-	outlen := C.int(0)
-	C._goboringcrypto_EVP_CipherUpdate(c.enc_ctx, (*C.uchar)(unsafe.Pointer(&dst[0])), &outlen, (*C.uchar)(unsafe.Pointer(&src[0])), C.int(aesBlockSize))
+	C.go_openssl_EVP_EncryptUpdate_wrapper(c.enc_ctx, base(dst), base(src), aesBlockSize)
```
ueno, line 95: Would it make sense to check the return value?

qmuntal: It does. We probably did not check the return because upstream Go is not checking in neither:

https://github.com/golang/go/blob/81efd7b347dd6d7f12fd49c6eee0274005734c71/src/crypto/internal/boring/aes.go#L186

BoringSSL [AES_ctr128_encrypt](https://commondatastorage.googleapis.com/chromium-boringssl-docs/aes.h.html#AES_ctr128_encrypt) does not return any value, so upstream has nothing to check. OpenSSL counterpart does return a status code, so we should be checking it.

openssl/aes.go
```diff
@@ -423,34 +352,49 @@ func (g *aesGCM) Seal(dst, nonce, plaintext, additionalData []byte) []byte {
 	if len(dst)+len(plaintext)+gcmTagSize < len(dst) {
 		panic("cipher: message too large for buffer")
 	}
+	if g.tls {
+		if len(additionalData) != gcmTlsAddSize {
+			panic("cipher: incorrect additional data length given to GCM TLS")
+		}
+		// BoringCrypto enforces strictly monotonically increasing explicit nonces
+		// and to fail after 2^64 - 1 keys as per FIPS 140-2 IG A.5,
+		// but OpenSSL does not perform this check, so it is implemented here.
```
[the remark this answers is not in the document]
qmuntal: We added this check to honor `crypto/tls` expectation that `boring.NewGCMTLS` validates key/iv pair uniqueness. If this is not implemented, then `boring.NewGCMTLS` and `aesCipher.NewGCM` will behave the same. Yet, no test is failing if this check is not implemented, as upstream is not testing this property.

FIPS 140-3 relaxed key-iv pair uniqueness constraints from SP 800-38D in some cases and added a couple of more options to be in compliance with this requirement, but I don't see where it downgrades this requirement to a recommendation. I reluctant to remove this check, as SP 800-38D Sec. 8 states that

> Compliance with this requirement is crucial to the security of GCM.

IMO doing nothing is not an option (I might be wrong here, correct me if that's the case), we either check it ourselves or force OpenSSL to do so, and the latter is not straightforward due to API incompatibilities.

About the lowered key usage limit in TLS 1.3, we could implement it in a latter PR so this current PR does not contain new features not available in either RedHat or Microsoft implementations.

derekparker:
@ueno Want to take another look at this when you get a chance?

openssl/aes.go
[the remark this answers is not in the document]
ueno: If the logic is aligned with `crypto/tls` then let's have it as it is; my concern was mainly that it might blur the boundary of FIPS module as the check shall be performed within a module.

ueno (approved):
Looks good to me!

PR #31, opened 2022-10-12
Add hash sign / hash verify

derekparker:
This API was removed during the transition to 1.19. This patch adds it back. It was originally requested because the FIPS lab insisted that hashing / signing (or verifying) happen in a single openssl call.

### DIFF
--- a/openssl/ecdsa.go
+++ b/openssl/ecdsa.go
@@ -10,14 +10,11 @@ package openssl
 // #include "goopenssl.h"
 import "C"
 import (
+	"crypto"
 	"errors"
 	"runtime"
 	"unsafe"
 )
-
-type ecdsaSignature struct {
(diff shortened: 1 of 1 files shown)


simo5:
LGTM

openssl/ecdsa.go
```diff
@@ -10,15 +10,12 @@ package openssl
 // #include "goopenssl.h"
 import "C"
 import (
+	"crypto"
 	"errors"
 	"runtime"
 	"unsafe"
 )
 
-type ecdsaSignature struct {
```
simo5, line 18: Was this a left over?
doesn't seem to be impacting the change otherwise

simo5 (approved):

ueno (approved):

PR #36, opened 2022-10-17
Restore original API for HashSign / HashVerify

derekparker:


dbenoit17 (approved):
LGTM

https://github.com/golang-fips/release/actions/runs/3273763766/jobs/5386400769

Issue #35, opened 2022-10-17
crypto segfaults on ubi7


dbenoit17:
TestBoringVerify segfaults on ubi7 when trying to call `_goboringcrypto_BN_lebin2bn()`

```
=== RUN   TestBoringVerify
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x0]

runtime stack:
runtime.throw({0x5ebfb2?, 0x7fcb3258412d?})
	/__w/openssl-fips/go/go/src/runtime/panic.go:1047 +0x5d fp=0x7ffcd36c1de0 sp=0x7ffcd36c1db0 pc=0x43bf3d
runtime.sigpanic()
	/__w/openssl-fips/go/go/src/runtime/signal_unix.go:819 +0x369 fp=0x7ffcd36c1e30 sp=0x7ffcd36c1de0 pc=0x452729

goroutine 7 [syscall]:
runtime.cgocall(0x59b990, 0xc000055c78)
	/__w/openssl-fips/go/go/src/runtime/cgocall.go:158 +0x5c fp=0xc000055c50 sp=0xc000055c18 pc=0x408afc
vendor/github.com/golang-fips/openssl-fips/openssl._Cfunc__goboringcrypto_BN_lebin2bn(0xc000128500, 0x80, 0x0)
	_cgo_gotypes.go:589 +0x4d fp=0xc000055c78 sp=0xc000055c50 pc=0x4d8dad
vendor/github.com/golang-fips/openssl-fips/openssl.bigToBN({0xc000128500?, 0xc000055ce0?, 0x51fc05?})
	/__w/openssl-fips/go/go/src/vendor/github.com/golang-fips/openssl-fips/openssl/openssl.go:197 +0x3b fp=0xc000055ca8 sp=0xc000055c78 pc=0x4df7fb
vendor/github.com/golang-fips/openssl-fips/openssl.NewPublicKeyRSA({0xc000128500, 0x10, 0x14}, {0xc00001e8e0, 0x1, 0x1})
	/__w/openssl-fips/go/go/src/vendor/github.com/golang-fips/openssl-fips/openssl/rsa.go:54 +0x6e fp=0xc000055cf0 sp=0xc000055ca8 pc=0x4e080e
crypto/rsa.boringPublicKey(0xc00002e950)
[...]

PR #37, opened 2022-10-18
Add fallback implementation of BN_lebin2bn and BN_bn2lebinpad

ueno:
As these functions are only available in OpenSSL 1.1.1 or later, this patch adds the fallback implementation, at the cost of additional memory allocation and byte reverse.

Fixes: #35
Signed-off-by: Daiki Ueno <dueno@redhat.com>

### DIFF
--- a/openssl/goopenssl.h
+++ b/openssl/goopenssl.h
@@ -448,7 +448,6 @@ typedef EC_KEY GO_EC_KEY;
 
 DEFINEFUNC(GO_EC_KEY *, EC_KEY_new, (void), ())
 DEFINEFUNC(GO_EC_KEY *, EC_KEY_new_by_curve_name, (int arg0), (arg0))
-DEFINEFUNC(int, EC_KEY_oct2key, (GO_EC_KEY *arg0, const unsigned char *arg1, size_t arg2, BN_CTX *arg3), (arg0, arg1, arg2, arg3))
 DEFINEFUNC(void, EC_KEY_free, (GO_EC_KEY * arg0), (arg0))
 DEFINEFUNC(const GO_EC_GROUP *, EC_KEY_get0_group, (const GO_EC_KEY *arg0), (arg0))
 DEFINEFUNC(int, EC_KEY_set_group, (GO_EC_KEY *arg0, const EC_GROUP *arg1), (arg0, arg1))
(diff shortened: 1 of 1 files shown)


ueno:
This partially fixes #35, though segfaults at some other symbol:

```console
=== RUN   TestBoringServerSignatureAndHash/0x50535357697468534841323536/fipstls
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x0]

runtime stack:
runtime.throw({0x774859?, 0x7f8b3648912d?})
	/opt/app-root/src/go/go/src/runtime/panic.go:1047 +0x5d fp=0x7f8b07184dc0 sp=0x7f8b07184d90 pc=0x43c5bd
runtime.sigpanic()
	/opt/app-root/src/go/go/src/runtime/signal_unix.go:819 +0x369 fp=0x7f8b07184e10 sp=0x7f8b07184dc0 pc=0x452e09

goroutine 118 [syscall]:
runtime.cgocall(0x6e6990, 0xc0000f91d8)
	/opt/app-root/src/go/go/src/runtime/cgocall.go:158 +0x5c fp=0xc0000f91b0 sp=0xc0000f9178 pc=0x408b7c
vendor/github.com/golang-fips/openssl-fips/openssl._Cfunc__goboringcrypto_EC_KEY_oct2key(0x7f8afc002d60, 0xc000274008, 0x41, 0x0)
	_cgo_gotypes.go:780 +0x4c fp=0xc0000f91d8 sp=0xc0000f91b0 pc=0x4f660c
vendor/github.com/golang-fips/openssl-fips/openssl.getPeerKey.func2(0x7f8afc002d60, {0xc000274008?, 0x41, 0x728fa0?})
	/opt/app-root/src/go/go/src/vendor/github.com/golang-fips/openssl-fips/openssl/ecdh.go:48 +0x76 fp=0xc0000f9218 sp=0xc0000f91d8 pc=0x4ff4b6
vendor/github.com/golang-fips/openssl-fips/openssl.getPeerKey(0x7f8afc002e10?, {0xc000274008, 0x41, 0x158})
	/opt/app-root/src/go/go/src/vendor/github.com/golang-fips/openssl-fips/openssl/ecdh.go:48 +0x8f fp=0xc0000f9270 sp=0xc0000f9218 pc=0x4ff36f
vendor/github.com/golang-fips/openssl-fips/openssl.SharedKeyECDH(0xc000208400?, {0xc000274008, 0x41, 0x158})
```

ueno:
The second commit ([Add fallback implementation of EC_KEY_oct2key](https://github.com/golang-fips/openssl-fips/pull/37/commits/58b805bd2e77ed1ac9341e867ddab65c92c28fbf)) should fix the segfault. The remaining failure is:
```console
=== RUN   TestBoringCertAlgs
    boring_test.go:481: 
        EC_KEY_new_by_curve_name failed
        openssl error(s):
        file: ec_curve.c
        line: 424
        function: 
        flags: 0
        error string: error:100AE081:lib(16):func(174):reason(129)
```

ueno:
The failing curve is P-224, which was [hobbled](https://src.fedoraproject.org/rpms/openssl/c/e443a79334446ac0dc14fdf7c062386f92bbc7a0?branch=e443a79334446ac0dc14fdf7c062386f92bbc7a0) back then.

derekparker (approved):
LGTM

dbenoit17:
@ueno I'm seeing one more segfault in ubi7 here https://github.com/golang-fips/release/actions/runs/3281732978/jobs/5404161719#step:14:234.  Could you take a look at that one?

ueno:
@dbenoit17 according to the log this is also related to unsupported P-224:
```console
    --- FAIL: TestSignAndVerify/P224 (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x572866]

goroutine 20 [running]:
[...]
crypto/ecdsa.testSignAndVerify(0xc000082680, {0x627f30?, 0x909cc0?})
	/__w/release/go/go/src/crypto/ecdsa/ecdsa_test.go:68 +0x8f
```
I suppose `ecdsa_test.go` needs to be adjusted to not exercise test cases with the curve.

qmuntal:
@ueno `BN_lebin2bn` and `BN_bn2lebinpad` are not exported in OpenSSL 1.0.2 headers but they do exist with a different name, and it is possibly to load them via `dlsym`. Try renaming name to `bn_lebin2bn` and `bn_bn2lebinpad`, it works in the MSFT bindings:

https://github.com/microsoft/go-crypto-openssl/blob/0e9fe572d632a5b387d301dcf2e9d5a735dbf318/openssl/openssl_funcs.h#L190-L192

ueno:
@qmuntal thanks for the information. I just checked whether these functions are present in ubi7 image, which is using 1.0.2k as
the base openssl version, but it seems they are not:
```console
sh-4.2$ rpm -qa openssl-libs
openssl-libs-1.0.2k-25.el7_9.x86_64
sh-4.2$ readelf -sW /usr/lib64/libcrypto.so | grep lebin
sh-4.2$ 
```
Looking at the git log, they were added in 1.0.2t, but I suppose in RHEL-7 only bug fixes had been backported from later patch releases. We can anyway try to load them and fallback to the suboptimal implementation in this PR.

derekparker:
> @qmuntal thanks for the information. I just checked whether these functions are present in ubi7 image, which is using 1.0.2k as the base openssl version, but it seems they are not:
> 
> ```
> $ podman run -ti --pull=always registry.access.redhat.com/ubi7/go-toolset:latest /bin/sh
> sh-4.2$ rpm -qa openssl-libs
> openssl-libs-1.0.2k-25.el7_9.x86_64
> sh-4.2$ readelf -sW /usr/lib64/libcrypto.so | grep lebin
> sh-4.2$ 
> ```
> 
> Looking at the git log, they were added in 1.0.2t, but I suppose in RHEL-7 only bug fixes had been backported from later patch releases. We can anyway try to load them and fallback to the suboptimal implementation in this PR.

Thanks for digging into this @ueno I think this sounds good to me. What are your thoughts @qmuntal ?

qmuntal:
> Thanks for digging into this @ueno I think this sounds good to me. What are your thoughts @qmuntal ?

Try to load those functions and fallback if they don't exit seems like the right approach. This way we don't slow down distros which do implement `bn_lebin2bn` and `bn_bn2lebinpad`.

ueno:
If we go with #42 (which always use big-endian), this PR would be unnecessary as the only user of those little-endian variants are in `openssl/bbig`.

dbenoit17 (approved):
LGTM

PR #41, opened 2022-11-13
Fix integer size confusion

ueno:
This makes the internal wrapper functions around OpenSSL functions consistently use `size_t` (or `size_t *`) where appropriate, instead of `unsigned int` (or `unsigned int *`).

The previous code had an issue on 64-bit big-endian systems where `sizeof(size_t)` is longer than `sizeof(unsigned int)` is 32, as the value returned through `unsigned int *` were truncated.

### DIFF
--- a/openssl/goopenssl.h
+++ b/openssl/goopenssl.h
@@ -414,7 +414,7 @@ typedef ECDSA_SIG GO_ECDSA_SIG;
 DEFINEFUNC(GO_ECDSA_SIG *, ECDSA_SIG_new, (void), ())
 DEFINEFUNC(void, ECDSA_SIG_free, (GO_ECDSA_SIG * arg0), (arg0))
 DEFINEFUNC(GO_ECDSA_SIG *, ECDSA_do_sign, (const uint8_t *arg0, size_t arg1, const GO_EC_KEY *arg2), (arg0, arg1, arg2))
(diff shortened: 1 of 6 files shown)


dbenoit17 (approved):
LGTM

openssl/goopenssl.h
```diff
@@ -218,10 +218,16 @@ DEFINEFUNC(const GO_EVP_MD *, EVP_sha512, (void), ())
 DEFINEFUNC(const GO_EVP_MD *, EVP_md_null, (void), ())
 #if OPENSSL_VERSION_NUMBER < 0x30000000L
 DEFINEFUNCINTERNAL(int, EVP_MD_type, (const GO_EVP_MD *arg0), (arg0))
+DEFINEFUNCINTERNAL(size_t, EVP_MD_size, (const GO_EVP_MD *arg0), (arg0))
```
qmuntal, line 221: `EVP_MD_size` and `EVP_MD_get_size` [docs](https://beta.openssl.org/docs/manmaster/man3/EVP_MD_get_size.html) say that these functions return an `int`, not a `size_t`.

ueno: Good catch, fixed.

qmuntal (approved):

PR #42, opened 2022-11-15
bbig: Reverse limbs depending on host endianness

ueno:
The conversion code between `BigInt` (= `[]uint`) and OpenSSL
`BIGNUM *` currently assumes that the uint array can be safely
accessed as `[]byte`.  While this is the case on little-endian
platforms, it requires reversal of each limb on big-endian
platforms (note that the ordering of limbs is always LSB first).

Signed-off-by: Daiki Ueno <dueno@redhat.com>

### DIFF
--- a/openssl/openssl.go
+++ b/openssl/openssl.go
@@ -14,6 +14,7 @@ package openssl
 */
 import "C"
 import (
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"math/bits"
@@ -45,9 +46,23 @@ var enabled = false
 // will still be done by OpenSSL.
 var strictFIPS = false
 
+var nativeEndian binary.ByteOrder
+
 func init() {
 	runtime.LockOSThread()
 	defer runtime.UnlockOSThread()
+
+	buf := [2]byte{}
+	*(*uint16)(unsafe.Pointer(&buf[0])) = uint16(0xABCD)
+
+	switch buf {
+	case [2]byte{0xCD, 0xAB}:
+		nativeEndian = binary.LittleEndian
+	case [2]byte{0xAB, 0xCD}:
+		nativeEndian = binary.BigEndian
+	default:
+		panic("Could not determine native endianness.")
+	}
 
 	// Check if we can `dlopen` OpenSSL
 	if C._goboringcrypto_DLOPEN_OPENSSL() == C.NULL {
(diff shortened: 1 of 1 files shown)


qmuntal:
The implementation seems OK, but I wonder if we could solve the host endianness issue by only using OpenSSL functions rather than doing a mix of OpenSSL and Go.

`math/big` is infamously known to be non-constant time, while OpenSSL's `BIGNUM` is mostly constant-time, particularly the functions that we use to convert big numbers from Go to OpenSSL, and viceversa. Source: https://github.com/openssl/openssl/pull/5254/commits/14a061dda3829a9d36aa01f993a5404089c28034.

AFAIU this PR makes the RSA and ECDSA bindings more vulnerable to timing attacks, although I'm not an export on this subject and I might be wrong.

Anyway, would it be possible to call the appropriate OpenSSL function based on the host endianness? I.e.:

- If host is little endian, use `BN_lebin2bn` and `BN_bn2lebinpad`.
- If host is big endian, use `BN_bin2bn` and `BN_bn2binpad`

ueno:
@qmuntal thanks, that's really a good point. I've managed to defer serializing/deserializing to the OpenSSL functions only with byteswap of each limb of BigInt. Could you take a look?

openssl/openssl.go
```diff
@@ -184,24 +199,47 @@ type fail string
...
 		return nil
 	}
 	return (*C.uint8_t)(unsafe.Pointer(&b[0]))
 }
 
-const wordBytes = bits.UintSize / 8
-
 func bigToBN(x BigInt) *C.GO_BIGNUM {
+	if nativeEndian == binary.BigEndian {
+		z := x
```
qmuntal, line 225: is this shallow copy necessary?

qmuntal:
> @qmuntal thanks, that's really a good point. I've managed to defer serializing/deserializing to the OpenSSL functions only with byteswap of each limb of BigInt. Could you take a look?

The new approach looks really nice!

openssl/openssl.go
ueno: It's not; removed.

qmuntal (approved):

openssl/openssl.go
ueno: Actually, we might need a (deep) copy of it, as this modifies the original array passed by reference. I guess it is not exposed through tests, just because the input variables are used only once.

derekparker (approved):
LGTM

PR #43, opened 2022-11-28
Implement RSA

qmuntal:
This PR implements the RSA API. The implementation is taken from the MSFT repo, which was heavily based on RedHat version.

Some comments for the reviewers:
- `SignRSAPKCS1v15` and `VerifyRSAPKCS1v15` function signatures are the ones defined by Boring, so they expect the message to be hashed and don't define the `msgIsHashed` parameter.
- I've added the functions `HashSignRSAPKCS1v15` and `HashVerifyRSAPKCS1v15` to cover the on-shot hash+sign use-case. Deciding which function to use is left to the backend layer.
- The `RSA_set0_factors`, `RSA_set0_crt_params`, and friends' fallbacks for OpenSSL 1.0.2 are implemented on the Go side instead of on the C side, following the logic that Go is safer and easier to maintain than C.
- I'm aware that #42 also affects the logic in this PR. I'll cherry pick it and create a follow-up PR once it is merged.

PR #44, opened 2022-12-01
Implement ECDSA

qmuntal:
⚠️ This PR will be rebased once #43 is merged. Please review #43 first.

This PR implements the ECDSA API. The implementation is taken from the MSFT repo, which was heavily based on RedHat version.

Some comments for the reviewers:
- I've added the functions `HashSignRSAPKCS1v15` and `HashVerifyRSAPKCS1v15` to cover the on-shot hash+sign use-case. Deciding which function to use is left to the backend layer.

### DIFF
--- a/openssl/ecdsa.go
+++ b/openssl/ecdsa.go
@@ -158,12 +158,6 @@ func newECKey(curve string, X, Y, D BigInt) (C.GO_EVP_PKEY_PTR, error) {
 	return pkey, nil
 }
 
-func bnFree(bn C.GO_BIGNUM_PTR) {
-	if bn != nil {
-		C.go_openssl_BN_free(bn)
-	}
-}
-
 func curveNID(curve string) (C.int, error) {
 	switch curve {
 	case "P-224":
--- a/openssl/ecdsa.go
+++ b/openssl/ecdsa.go
@@ -1,0 +1,170 @@
+//go:build linux && !android
+// +build linux,!android
+
+package openssl
+
+// #include "goopenssl.h"
+import "C"
+import (
+	"errors"
+	"runtime"
+)
+
+type PrivateKeyECDSA struct {
+	// _pkey MUST NOT be accessed directly. Instead, use the withKey method.
+	_pkey C.GO_EVP_PKEY_PTR
+}
+
+func (k *PrivateKeyECDSA) finalize() {
+	C.go_openssl_EVP_PKEY_free(k._pkey)
+}
+
+func (k *PrivateKeyECDSA) withKey(f func(C.GO_EVP_PKEY_PTR) C.int) C.int {
+	defer runtime.KeepAlive(k)
+	return f(k._pkey)
+}
+
+type PublicKeyECDSA struct {
+	// _pkey MUST NOT be accessed directly. Instead, use the withKey method.
+	_pkey C.GO_EVP_PKEY_PTR
+}
+
+func (k *PublicKeyECDSA) finalize() {
+	C.go_openssl_EVP_PKEY_free(k._pkey)
+}
+
+func (k *PublicKeyECDSA) withKey(f func(C.GO_EVP_PKEY_PTR) C.int) C.int {
+	defer runtime.KeepAlive(k)
+	return f(k._pkey)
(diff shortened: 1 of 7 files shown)


openssl/ecdsa.go
```diff
@@ -0,0 +1,178 @@
...
+		return nil, err
+	}
+	// Create a new EC_KEY for the given curve.
+	key := C.go_openssl_EC_KEY_new_by_curve_name(nid)
+	if key == nil {
+		return nil, newOpenSSLError("EC_KEY_new_by_curve_name failed")
+	}
+	defer C.go_openssl_EC_KEY_free(key)
+
+	// Convert X, Y, and D coordinates to OpenSSL format.
```
dagood, line 129: Maybe mention here that D is optional so the `D` check is more obvious later in `bx == nil || by == nil || (D != nil && bd == nil)`.

```suggestion
	// Convert X, Y, and D coordinates to OpenSSL format. D is optional and may be nil.
```

openssl/ecdsa.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+func VerifyECDSA(pub *PublicKeyECDSA, hash []byte, sig []byte) bool {
+	return evpVerify(pub.withKey, 0, 0, 0, sig, hash) == nil
+}
+
+func HashVerifyECDSA(pub *PublicKeyECDSA, h crypto.Hash, msg, sig []byte) bool {
+	return evpHashVerify(pub.withKey, h, msg, sig) == nil
+}
+func newECKey(curve string, X, Y, D BigInt) (C.GO_EVP_PKEY_PTR, error) {
```
dagood, line 117: Missing newline

```suggestion
}

func newECKey(curve string, X, Y, D BigInt) (C.GO_EVP_PKEY_PTR, error) {
```

openssl/ecdsa.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	// Create the EVP_PKEY and assign it the EC_KEY.
+	pkey := C.go_openssl_EVP_PKEY_new()
+	if pkey == nil {
+		return nil, newOpenSSLError("EVP_PKEY_new failed")
+	}
+	if C.go_openssl_EVP_PKEY_set1_EC_KEY(pkey, key) != 1 {
+		C.go_openssl_EVP_PKEY_free(pkey)
+		return nil, newOpenSSLError("EVP_PKEY_assign failed")
```
dagood, line 155:
```suggestion
		return nil, newOpenSSLError("EVP_PKEY_set1_EC_KEY failed")
```

openssl/ecdsa.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	return pkey, nil
+}
+
+func bnFree(bn C.GO_BIGNUM_PTR) {
```
dagood, line 160: Does this func really belong here? Or `openssl.go` with the other `bn` funcs?

qmuntal: Good catch. Done.

dagood (approved):

ueno (approved):

openssl/ecdsa.go
```diff
@@ -0,0 +1,173 @@
...
+	C.go_openssl_EVP_PKEY_free(k._pkey)
+}
+
+func (k *PublicKeyECDSA) withKey(f func(C.GO_EVP_PKEY_PTR) C.int) C.int {
+	defer runtime.KeepAlive(k)
+	return f(k._pkey)
+}
+
+var errUnknownCurve = errors.New("openssl: unknown elliptic curve")
+var errUnsupportedCurve = errors.New("openssl: unsupported elliptic curve")
```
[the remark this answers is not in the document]
qmuntal: Good catch. This is a leftover from when P224 was not implemented. If Fedora/RHEL don't support a curve that can be handled in the backend layer, or we can try to implement (in a future PR) a way to prove which algorithms are supported by the underlying crypto.

qmuntal:
@derekparker waiting for your review 😄

derekparker (approved):
LGTM

PR #46, opened 2023-01-19
Add SupportsHKDF()

dbenoit17:
This commit introduces a function that returns true or false depending on whether the OpenSSL version is expected to support HKDF. It is expected to be used by the crypto/tls library to help determine whether TLS version 1.3 can be used in FIPS mode.

### DIFF
--- a/openssl/hkdf.go
+++ b/openssl/hkdf.go
@@ -19,6 +19,10 @@ type hkdf struct {
 	ctx *C.GO_EVP_PKEY_CTX
 }
 
+func SupportsHKDF() bool {
+	return openSSLVersion() >= OPENSSL_VERSION_1_1_1
+}
+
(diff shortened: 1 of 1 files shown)


openssl/hkdf.go
```diff
@@ -19,6 +19,13 @@ type hkdf struct {
 	ctx *C.GO_EVP_PKEY_CTX
 }
 
+func SupportsHKDF() bool {
+	if openSSLVersion() < OPENSSL_VERSION_1_1_1 {
```
derekparker, line 23: We can eliminate the branch and just write `return openSSLVersion() >= OPENSSL_VERSION_1_1_1`.

derekparker (approved):
LGTM

ueno (approved):

PR #47, opened 2023-01-24
Fix memory leak in GenerateKeyRSA

qmuntal:
`GenerateKeyRSA` calls `EVP_PKEY_get1_RSA` without freeing the returned pointer, causing a leak.

### DIFF
--- a/openssl/rsa.go
+++ b/openssl/rsa.go
@@ -27,6 +27,7 @@ func GenerateKeyRSA(bits int) (N, E, D, P, Q, Dp, Dq, Qinv BigInt, err error) {
 	if key == nil {
 		return bad(newOpenSSLError("EVP_PKEY_get1_RSA failed"))
 	}
+	defer C.go_openssl_RSA_free(key)
 	N, E, D = rsaGetKey(key)
 	P, Q = rsaGetFactors(key)
 	Dp, Dq, Qinv = rsaGetCRTParams(key)


dagood (approved):

openssl/rsa.go
```diff
@@ -27,6 +27,7 @@ func GenerateKeyRSA(bits int) (N, E, D, P, Q, Dp, Dq, Qinv BigInt, err error) {
 	if key == nil {
 		return bad(newOpenSSLError("EVP_PKEY_get1_RSA failed"))
 	}
+	defer C.go_openssl_RSA_free(key)
```
ueno, line 30: I was wondering if could use `get0_RSA` instead of `get1_RSA` above, to not increase the reference count?

ueno (approved):

openssl/rsa.go
qmuntal: Unfortunately  `get0_RSA` is not available on OpenSSL 1.0.2, so it's simpler to just use `get1_RSA`.

qmuntal: Thinking it twice, OpenSSL 1.0.2 does have `EVP_PKEY_get0`, which gets the raw key as an untyped void pointer. We could use it to bridge the lack of `EVP_PKEY_get0_RVA`. Will do that on a follow up PR.

PR #49, opened 2023-02-09
Implement HMAC using EVP_MAC_CTX on OpenSSL 3

qmuntal:
This PR updates the HMAC implementation to use `EVP_MAC_CTX` on OpenSSL 3 instead of the deprecated `HMAC_CTX`.

While here, I've changed how OpenSSL 1 resets the context to avoid having to keep the key around. [HMAC_Init_ex](https://www.openssl.org/docs/man1.0.2/man3/HMAC_Init_ex.html) reuses the same key and hash function if they are not passed, as `Reset()` is not changing those parameters, there is no need to pass them.

> HMAC_Init_ex() initializes or reuses a HMAC_CTX structure to use the hash function evp_md and key key. If both are NULL (or evp_md is the same as the previous digest used by ctx and key is NULL) the existing key is reused. ctx must have been created with HMAC_CTX_new() before the first use of an HMAC_CTX in this function. N.B. HMAC_Init() had this undocumented behaviour in previous versions of OpenSSL - failure to switch to HMAC_Init_ex() in programs that expect it will cause them to stop working. 

Unfortunately the same optimization can't be used on OpenSSL 3 due to this bug: https://github.com/openssl/openssl/issues/17811.

### DIFF
--- a/openssl/ecdh.go
+++ b/openssl/ecdh.go
@@ -11,12 +11,6 @@ import (
 	"unsafe"
 )
 
-var (
-	paramPubKey  = C.CString("pub")
-	paramPrivKey = C.CString("priv")
-	paramGroup   = C.CString("group")
-)
-
 type PublicKeyECDH struct {
 	_pkey C.GO_EVP_PKEY_PTR
 	bytes []byte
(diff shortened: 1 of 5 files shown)


derekparker (approved):
LGTM, will wait for @ueno to review as well.

dagood (approved):

qmuntal:
Friendly remainder to review this PR

simo5 (approved):
LGTM

PR #50, opened 2023-02-17
Remove wrapper around BN_free

ueno:
The wrapper `bnFree` was there only to guard `BN_free` from a potentially
NULL argument, while `BN_free` actually takes care of it as it follows
the same semantics of free.


qmuntal (approved):
I wrapped the `BN_free` call to avoid the overhead of an unnecessary cgo call in case the bignum was null, but the speed-up is almost noticiable so I'm ok removing the wrapper if it adds more confusions.

ueno:
Thank you for the background. I'm merging this for now, but we could revisit later when there is a performance hit.

PR #53, opened 2023-02-22
Add missing methods for ECDH keys

derekparker:


ueno (approved):
Yeah, thanks for spotting this!

PR #54, opened 2023-02-23
Fix missing symbols between openssl versions

derekparker:


openssl/goopenssl.h
```diff
@@ -1054,19 +1062,43 @@ DEFINEFUNC(int, EVP_PKEY_keygen_init, (GO_EVP_PKEY_CTX *ctx), (ctx))
...
 };
+enum {
+	GO_POINT_CONVERSION_UNCOMPRESSED = POINT_CONVERSION_UNCOMPRESSED,
+};
+#endif
 
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L
+#if OPENSSL_VERSION_NUMBER < 0x30000000L
+typedef struct ossl_param_st GO_OSSL_PARAM;
+struct ossl_param_st {
```
ueno, line 1081: It looks a bit cumbersome to expose the OpenSSL 3-only interface to make `params.go` compile, though it is practically unused when an earlier version of OpenSSL is detected at run time. Perhaps, we could port the code using OSSL_PARAM into C and only expose minimal set of functions?

ueno (approved):

openssl/goopenssl.h
ueno: I'm not sure if it is feasible but perhaps we could use a build tag for conditional compilation, e.g., openssl10, openssl11, and openssl3, so that
- `newECDHPkey`  is defined differently depending on the build tag
- `params.go` (used by `newECDHPkey` implementation for OpenSSL 3) is only compiled with `+build openssl3`

derekparker: I think a build tag could work, however that could complicate the RPM builds a bit so I would like @dbenoit17 input on that.

ueno:
Superseded by #55.

PR #55, opened 2023-03-06
Avoid exposing OSSL_PARAM_* through CGO

ueno:
As we don't make use of build conditionals, params.go needs to be
compiled against all OpenSSL versions and that requires stub
definitions of the 3.0-only API to be exposed from CGO.  This
minimizes the CGO interface to use the API only internally in the C
level.


derekparker:
I know this is still marked as draft but initially it LGTM. I really like how this cleans things up a lot and has the added benefit of reducing the overall number of CGO calls we make making it more efficient overall.

ueno:
There are still some rough edges, but I confirm `go test` passes with both OpenSSL 3.0 and 1.1.

dbenoit17 (approved):

PR #56, opened 2023-03-06
ecdsa: don't use EC_KEY functions on OpenSSL 3

qmuntal:
This PR updates the ECDSA implementation so that it doesn't use the now-deprectaed EC_KEY_* functions, which don't work well with the FIPS provider.


openssl/evp.go
```diff
@@ -393,3 +393,34 @@ func newEvpFromParams(id C.int, selection C.int, params []C.OSSL_PARAM) (C.GO_EV
...
+	bytes := make([]byte, n)
+	n = C.go_openssl_EC_POINT_point2oct(group, pt, C.GO_POINT_CONVERSION_UNCOMPRESSED, base(bytes), n, nil)
+	if n == 0 {
+		return nil, newOpenSSLError("EC_POINT_point2oct")
+	}
+	return bytes, nil
+}
+
+// deriveAndEncode calls derive to generate a public key point and then encodes it.
+func deriveAndEncode(nid C.int, derive func(group C.GO_EC_GROUP_PTR) (C.GO_EC_POINT_PTR, error)) ([]byte, error) {
```
simo5, line 414: The term "derive" confuses me a little bit here.

Do I read it right that this is referring to just extracting a public key from the private key?

Or is this ECDH or other kind of actual key derivation ?

qmuntal: Good point. This function generates and encodes a public key, it doesn't derive anything. Will rename it to `generateAndEncodePublicKey`.

openssl/params.go
```diff
@@ -99,3 +101,14 @@ func (pb *paramsBuilder) addBigNumber(key *C.char, v []byte) error {
 	pb.add(key, C.GO_OSSL_PARAM_UNSIGNED_INTEGER, unsafe.Pointer(cbytes), C.size_t(len(v)))
 	return nil
 }
+
+// addBN adds a parameter of type OSSL_PARAM_UNSIGNED_INTEGER.
+func (pb *paramsBuilder) addBN(key *C.char, v C.GO_BIGNUM_PTR) error {
+	size := (C.go_openssl_BN_num_bits(v) + 7) / 8
+	x := C.malloc(C.size_t(size))
+	if C.go_openssl_BN_bn2nativepad(v, (*C.uchar)(x), size) == 0 {
+		return newOpenSSLError("BN_bn2nativepad")
```
dagood, line 110: Would this be a memory leak in the error case?

qmuntal: Good catch.

openssl/evp.go
```diff
@@ -393,3 +393,34 @@ func newEvpFromParams(id C.int, selection C.int, params []C.OSSL_PARAM) (C.GO_EV
...
+	bytes := make([]byte, n)
+	n = C.go_openssl_EC_POINT_point2oct(group, pt, C.GO_POINT_CONVERSION_UNCOMPRESSED, base(bytes), n, nil)
+	if n == 0 {
+		return nil, newOpenSSLError("EC_POINT_point2oct")
+	}
+	return bytes, nil
+}
+
+// generateAndEncodePublicKey calls newPubKeyPointFn to generate a public key point and then encodes it.
+func generateAndEncodePublicKey(nid C.int, newPubKeyPointFn func(group C.GO_EC_GROUP_PTR) (C.GO_EC_POINT_PTR, error)) ([]byte, error) {
```
ueno, line 414: Is this the right place to add these couple of functions? They are not using `EVP_*` functions nor being used by `evp.go`.

openssl/params.go
```diff
@@ -99,3 +101,15 @@ func (pb *paramsBuilder) addBigNumber(key *C.char, v []byte) error {
 	pb.add(key, C.GO_OSSL_PARAM_UNSIGNED_INTEGER, unsafe.Pointer(cbytes), C.size_t(len(v)))
 	return nil
 }
+
+// addBN adds a parameter of type OSSL_PARAM_UNSIGNED_INTEGER.
+func (pb *paramsBuilder) addBN(key *C.char, v C.GO_BIGNUM_PTR) error {
```
ueno, line 106: Now we have `addBigNumber` and `addBN`, which looks a bit misleading; perhaps we could rename the former?

By the way, in #55 I use `OSSL_PARAM_BLD_*` in the C level, and `OSSL_PARAM_BLD_push_BN` seems to take care of the endianness conversion.

ueno (approved):

openssl/params.go
qmuntal: Thanks for the suggestion @ueno. I've realized that `paramsBuilder` was causing more harm than good, so I've removed it and changed all usages to the `OSSL_PARAM_BLD` interface.

openssl/evp.go
qmuntal: Agree. I've moved these functions, and a couple others, to the new `ec.go` file.

dagood (approved):

PR #57, opened 2023-03-31
Stop using RSA_* functions for signatures

ueno:
For creating and verifying PKCS#1 v1.5 signatures in a pre-hashed manner, we used the legacy `RSA_sign` and `RSA_verify` functions, which bypass the system-wide disablement of SHA-1 and shorter RSA key length usage inconsistently with the OpenSSL 3.0 default on RHEL.

This switches to using our `_goboringcrypto_EVP_{sign,verify}_raw`, which internally use `EVP_PKEY_` functions.

### DIFF
--- a/openssl/goopenssl.h
+++ b/openssl/goopenssl.h
@@ -575,9 +575,6 @@ int _goboringcrypto_ECDSA_verify(EVP_MD *md, const uint8_t *arg1, size_t arg2, c
 
 #include <openssl/rsa.h>
 
-// Note: order of struct fields here is unchecked.
-typedef BN_GENCB GO_BN_GENCB;
-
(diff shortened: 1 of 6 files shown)


simo5 (approved):
Looks good in general but I wonder why the change in behavior for msgIsHashed cases, I think we wanted to enforce hashing in the FIPS boundary in the past?

ueno:
Afaik we later had to revert that enforcement as the [crypto/rsa.SignPKCS1v15](https://pkg.go.dev/crypto/rsa#SignPKCS1v15) takes hashed bytes and there are still some use of PKCS#1 v1.5 signatures in applications.

derekparker (approved):
LGTM

PR #58, opened 2023-04-10
Stop using deprecated EC functions for ECDSA

ueno:
This removes the usage of EC_KEY_* functions for cryptographic operations mainly to enforce FIPS restrictions; for accessors, some of them are still used. The first two commits are from #57.


derekparker (approved):
LGTM

PR #59, opened 2023-04-19
ecdh, ecdsa: Make sure to clear private BIGNUM values

ueno:
This adds zeroization of critical security parameters in ECDH and ECDSA operations, to comply with the FIPS 140-3 requirements.


qmuntal (approved):
Good catch!

PR #61, opened 2023-04-20
hkdf: Implement HKDF

ueno:
This implements the same interface as `x/crypto/hkdf`, using the OpenSSL API supported since 1.1.1.


openssl/hkdf.go
```diff
@@ -0,0 +1,159 @@
...
+		return nil, errors.New("unsupported hash function")
+	}
+
+	ctx := C.go_openssl_EVP_PKEY_CTX_new_id(C.GO_EVP_PKEY_HKDF, nil)
+	if ctx == nil {
+		return nil, newOpenSSLError("EVP_PKEY_CTX_new_id")
+	}
+	defer func() {
+		C.go_openssl_EVP_PKEY_CTX_free(ctx)
+	}()
```
qmuntal, line 39:
```suggestion
	defer C.go_openssl_EVP_PKEY_CTX_free(ctx)
```

openssl/hkdf.go
```diff
@@ -0,0 +1,159 @@
...
+		}
+		if C.go_openssl_EVP_PKEY_CTX_ctrl(ctx, -1, C.GO1_EVP_PKEY_OP_DERIVE,
+			C.GO_EVP_PKEY_CTRL_HKDF_MD,
+			0, unsafe.Pointer(md)) != 1 {
+			return nil, newOpenSSLError("EVP_PKEY_CTX_set_hkdf_md")
+		}
+	}
+
+	c := &hkdf{ctx: ctx}
+	ctx = nil
```
qmuntal, line 66: is this null assignment necessari?

qmuntal (changes requested):
Thanks!

openssl/hkdf.go
qmuntal: oh, I see. It makes sense.

qmuntal (approved):

PR #64, opened 2023-04-25
sha: add WriteString and WriteByte method

qmuntal:
This PR implements WriteString and WriteByte for all supported hashers. These methods will be required by the boring API in go1.21.

See [CL 483816](https://go-review.googlesource.com/c/go/+/483816), [483815](https://go-review.googlesource.com/c/go/+/483815), and [CL 481478](https://go-review.googlesource.com/c/go/+/481478).

### DIFF
--- a/openssl/sha_test.go
+++ b/openssl/sha_test.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"encoding"
 	"hash"
+	"io"
 	"testing"
 
 	"github.com/golang-fips/openssl-fips/openssl"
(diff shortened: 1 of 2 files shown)


ueno (approved):

openssl/sha_test.go
```diff
@@ -61,6 +62,24 @@ func TestSha(t *testing.T) {
...
+			for i := 0; i < len(msg); i++ {
+				bw.WriteByte(msg[i])
+			}
+			h.Reset()
+			sum = h.Sum(nil)
+			if !bytes.Equal(sum, initSum) {
+				t.Errorf("got:%x want:%x", sum, initSum)
+			}
+
+			h.(io.StringWriter).WriteString(string(msg))
```
dagood, line 82: The test pattern seems off by one... these don't seem like they should be one after another at the start of this added chunk:

```go
			h.Reset()
			sum = h.Sum(nil)
			if !bytes.Equal(sum, initSum) {
				t.Errorf("got:%x want:%x", sum, initSum)
			}

			h.Reset()
			sum = h.Sum(nil)
			if !bytes.Equal(sum, initSum) {
				t.Errorf("got:%x want:%x", sum, initSum)
			}
```

And following the general pattern of this test, after this `WriteString` line, I would expect another `Reset` and `Sum` check.

dagood (approved):
Other than the test comment, LGTM.

openssl/sha_test.go
qmuntal: Good catch, bad copy paste from `go-crypto-openssl`.

PR #66, opened 2023-05-29
Remove Init fallback logic and provide CheckVersion instead

qmuntal:
This PR removes the fallback logic of `openssl.Init("")`, which would try to load OpenSSL versions from a list of well-known shared library sufixes. The reasoning to do so is that the fallback process assumes all Go forks using `openssl-fips` have the same OpenSSL requirements and priorities, and that the list is static and hardly changes. For example, the MSFT Go fork wants to prioritize FIPS-enabled OpenSSL versions, even if it's not the highest available version.

A better approach, implemented also in this PR, is to expose the necessary functions so the caller always calls `openssl.Init` with the right version. To do so, I've implemented `openssl.CheckVersion`, which checks if a given version exists and is FIPS capable.

Downstream users could implement the initialization process as follows:

```go
func init() {
	var knownVersions = []string{"3", "1.1", "1.0"}
	fallbackIdx := -1
	for i, v := range knownVersions {
		exists, fips := openssl.CheckVersion(v)
		if exists && fips {
			_ = openssl.Init(v)
			return
		}
		if exists && fallbackIdx == -1 {
			fallbackIdx = i
		}
	}
	if fallbackIdx != -1 {
		_ = openssl.Init(knownVersions[fallbackIdx])
	}
}
```

Additionally, I've updated the `openssl.SetFips` so it always tries to load a provider named `fips` when `enabled` is true. It previously implemented some logic to allow other provider names if the OpenSSL config file was properly configured. I no longer think this is a good thing to support, first because we are not testing that code path, but more importantly because no other framework (i.e. [nodejs](https://github.com/nodejs/node/blob/607c8f4eacd5dded97a33f0da49da0aed7404187/src/crypto/crypto_util.cc#L120-L138), [python](https://github.com/pyca/cryptography/blob/5efbc110905230f771d5bde918c059cd3d029e22/src/cryptography/hazmat/bindings/openssl/binding.py#L86-L100), and [ruby](https://github.com/ruby/ruby/blob/d5ef373b1194bac64784ae316d125d7a2cf1988a/ext/openssl/ossl.c#L447-L459)) supports this use case. Being this in the critical security path, I would rather be cautious and don't support things that aren't fully understood.

### DIFF
--- a/openssl/init.go
+++ b/openssl/init.go
@@ -18,7 +18,6 @@ import (
 // See Init() for details about version.
 func opensslInit(version string) (major, minor, patch int, err error) {
 	// Load the OpenSSL shared library using dlopen.
-	// If version is specified try to load it or error out.
 	handle := dlopen(version)
 	if handle == nil {
 		errstr := C.GoString(C.dlerror())
--- a/openssl/init.go
+++ b/openssl/init.go
@@ -11,16 +11,6 @@ import (
 	"unsafe"
 )
 
-// knownVersions is a list of supported and well-known libcrypto.so suffixes in decreasing version order.
-//
-// FreeBSD library version numbering does not directly align to the version of OpenSSL.
-// Its preferred search order is 11 -> 111.
-//
-// Some distributions use 1.0.0 and others (such as Debian) 1.0.2 to refer to the same OpenSSL 1.0.2 version.
-//
-// Fedora derived distros use different naming for the version 1.0.x.
-var knownVersions = [...]string{"3", "1.1", "11", "111", "1.0.2", "1.0.0", "10"}
-
 // opensslInit loads and initialize OpenSSL.
 // If successful, it returns the major and minor OpenSSL version
 // as reported by the OpenSSL API.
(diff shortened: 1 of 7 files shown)


openssl/init.go
```diff
@@ -11,26 +11,18 @@ import (
...
 // If successful, it returns the major and minor OpenSSL version
 // as reported by the OpenSSL API.
 //
 // See Init() for details about version.
 func opensslInit(version string) (major, minor, patch int, err error) {
 	// Load the OpenSSL shared library using dlopen.
-	handle, err := loadLibrary(version)
-	if err != nil {
-		return 0, 0, 0, err
+	// If version is specified try to load it or error out.
```
ueno, line 21: `version` is always specified, right?

openssl/openssl_test.go
```diff
@@ -28,3 +28,14 @@ func TestMain(m *testing.M) {
 	openssl.CheckLeaks()
 	os.Exit(status)
 }
+
+func TestCheckVersion(t *testing.T) {
+	v := os.Getenv("GO_OPENSSL_VERSION_OVERRIDE")
```
ueno, line 33: Am I right that this envvar should always set when running tests? Perhaps it makes sense to log something if `v` is ""?

ueno (approved):
Looks like a sensible approach to me; only minor nits inlined.

openssl/init.go
qmuntal: true, will remove that line.

openssl/openssl_test.go
qmuntal: Good point. I've improved `TestMain` so it tries to find an available OpenSSL version in case `GO_OPENSSL_VERSION_OVERRIDE` is not set.

PR #68, opened 2023-05-30
run runtime.GC() a few times before checking for leaks

### LINKED ISSUE #67: Memory leak detected in TestEncryptDecryptOAEP
Found while running CI test for #66.

Logs: https://github.com/golang-fips/openssl-fips/actions/runs/5121917944/jobs/9210281419#step:9:333
 
Can reproduce locally running `for run in {1..50}; do go test ./openssl -gcflags=all=-d=checkptr -asan -run ^TestEncryptDecryptOAEP$ -count 20 -v || break; done` (the leak is hard to get detected). Will investigate.

qmuntal:
Some objects are still alive when `openssl.CheckLeaks()` is called. Run `runtime.GC()` a few times more to avoid false positives while detecting memory leaks.

I've found this while investigating #67. Thanks to building OpenSSL in debug mode I could track down all the leaks to objects still not garbage collected,

Closes #67

### DIFF
--- a/openssl/openssl_test.go
+++ b/openssl/openssl_test.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"runtime"
 	"testing"
+	"time"
 
 	"github.com/golang-fips/openssl-fips/openssl"
 )


derekparker:
Does this deterministically free all remaining live objects? What was still live after the first GC? I suppose there's no harm in calling GC multiple times vs a sleep and single call to `runtime.GC`, but I'm curious if we could wait for something more deterministic to ensure there's no live objects before triggering manual GC.

dagood:
I was looking at this too--it seems like there's got to be a better way. 😄 I saw that in golang/go tests, repeated `runtime.GC()` calls are used: https://grep.app/search?q=runtime.GC%28%29&filter[lang][0]=Go&filter[repo][0]=golang/go, but I wasn't able to find any justifications.

I think an alternative would be one call to `runtime.GC()` to make finalizers start running, then sleep some amount of time until it's all done. As far as I can tell, extra `runtime.GC()` calls might just be acting as a sleep, so maybe this is actually equivalent?

There doesn't appear to be any way to tell if there are any arbitrary finalizers in the queue. https://github.com/golang/go/blob/993707a9d62d7541afe47d13caf5e3555e317702/test/fixedbugs/issue24491b.go uses a channel to signal from the finalizer to the outer code when the finalizer runs--but I don't see how we could reasonably instrument all the code this way just for testing purposes.

scripts/openssl.sh
```diff
@@ -48,7 +48,9 @@ rm -rf "openssl-$version"
 mv "openssl-$tag" "openssl-$version"
 
 cd "openssl-$version"
-./config $config
+# -d makes a debug build which helps with debugging memory issues and
+# other problems. It's not necessary for normal use.
+./config -d $config
```
dagood, line 53: I wonder if this should be done as an optional flag on the script in case debug mode changes behavior, ending up making us miss some edge case. On the other hand, I think this might enable debug assertions and let us catch other issues earlier (e.g. we violate some invariant).

Or would it be justified to add another variable to the build matrix and run both ways?

dagood (approved):
In my opinion given this reduces test flakiness in practice, it's ok to accept repeated `runtime.GC()` before fully understanding the reason or finding a better way.

I'm less sure about the consequences of `-d`. Maybe it's a minimal difference and doesn't matter as far as testing these bindings goes.

qmuntal:
> Does this deterministically free all remaining live objects? What was still live after the first GC? I suppose there's no harm in calling GC multiple times vs a sleep and single call to runtime.GC, but I'm curious if we could wait for something more deterministic to ensure there's no live objects before triggering manual GC.

We could call [runtime.ReadMemStats](https://pkg.go.dev/runtime#ReadMemStats) and check for `MemStatus.HeapAlloc`, but that's tricky, because it also includes runtime allocations.

> runtime.GC() calls are used: [https://grep.app/search?q=runtime.GC%28%29&filter[lang][0]=Go&filter[repo][0]=golang/go](https://grep.app/search?q=runtime.GC%28%29&filter%5Blang%5D%5B0%5D=Go&filter%5Brepo%5D%5B0%5D=golang/go), but I wasn't able to find any justifications.

There are a couple of places where they mention that multiple `runtime.GC()` cycles are needed to free live objects:

- https://github.com/golang/go/blob/094c75219a160be1667c9757f363ffad8926632b/src/cmd/compile/internal/noder/unified.go#L275-L284
- https://github.com/golang/go/blob/094c75219a160be1667c9757f363ffad8926632b/src/os/exec/exec_test.go#L91-L94

> I think an alternative would be one call to runtime.GC() to make finalizers start running, then sleep some amount of time until it's all done. As far as I can tell, extra runtime.GC() calls might just be acting as a sleep, so maybe this is actually equivalent?

`runtime.GC()` docs says this: `GC runs a garbage collection and blocks the caller until the garbage collection is complete.` So I don't think sleeping would help here.

scripts/openssl.sh
qmuntal: The chances to miss an edge-case by using a debug build are very low, but they do exist. Yet, I don't think it matters in practice as downstream users won't be using the OpenSSL libraries we use for testing, either because the patch version is not the same, or because they patch OpenSSL, or because the build configuration is different.

With this in mind, I think that it's up to the Microsoft and RedHat Go forks to test on environments that would match better their end users.

derekparker (approved):
LGTM

dagood:
> `runtime.GC()` docs says this: `GC runs a garbage collection and blocks the caller until the garbage collection is complete.` So I don't think sleeping would help here.

The GC completes, but the finalizer running isn't part of GC, it's just added to a queue to run later on a finalizer-running goroutine. Here's a test--it's easier than I thought it would be to make a pretty clear example after some liberal use of `time.Sleep`. 😄

https://go.dev/play/p/0_R8Vc1lvdW

```
	fmt.Println("main - Setting up.")
	setup()
	runtime.GC()
	runtime.GC()
	runtime.GC()
	runtime.GC()
	runtime.GC()
	fmt.Println("main - Ok, now I'll wait.")
	time.Sleep(time.Second * 2)
	fmt.Println("main - Done waiting.")
```

With `setup()` making a garbage struct with a one-second wait in its finalizer, the finalizer finishes running during the two-second sleep in this main method.

qmuntal:
> The GC completes, but the finalizer running isn't part of GC, it's just added to a queue to run later on a finalizer-running goroutine. Here's a test--it's easier than I thought it would be to make a pretty clear example after some liberal use of time.Sleep. 😄

Wow, thanks for digging into this and proving me wrong, I've learned something today 😄

> As far as openssl-fips goes, I think this might mean that both repeated calls to GC and sleeps might be necessary to be even more sure all the finalizers are queued up and have had time to run. Since each GC does a sweep, that seems to me to provide some built-in sleep, which seems good enough to fix the issue.

I'll add a small sleep on our side too, just in case GC implementation changes to not sweep.

PR #69, opened 2023-06-01
Allow hash.Hash for OAEP and MGF1 to be specified independently

qmuntal:
Go 1.20 allows hash.Hash for RSA-OAEP and MGF1 to be specified independently: https://go-review.googlesource.com/c/go/+/418874.

To do so Go broke the Boring API by adding a new optional parameter to `EncryptRSAOAEP` and `DecryptRSAOAEP`. We can do that here too, as this branch is still experimental and doesn't have to maintain backwards compatibility.

This is a partial port of https://github.com/microsoft/go-crypto-openssl/pull/42.


ueno (approved):
Looks good to me.

dagood (approved):

PR #75, opened 2023-06-14
Cache and pre-fetch EVP_MD objects

qmuntal:
This PR implements the following OpenSSL performance guidelines:

- EVP_MD objects should be fetched just once and used for multiple calls to other operations such as EVP_DigestInit_ex().
- (For OpenSSL 3) If you perform the same operation many times then it is recommended to use ["Explicit fetching"](https://www.openssl.org/docs/man3.0/man7/crypto.html#Explicit-fetching) to prefetch an algorithm once initially, and then pass this created object to any operations that are currently using ["Implicit fetching"](https://www.openssl.org/docs/man3.0/man7/crypto.html#Implicit-fetching).

See ["Performance" in crypto(7)](https://www.openssl.org/docs/man3.0/man7/crypto.html) for further information.

### DIFF
--- a/evp.go
+++ b/evp.go
@@ -14,6 +14,7 @@ import (
 	"unsafe"
 )
 
+// cacheMD is a cache of crypto.Hash to GO_EVP_MD_PTR.
 var cacheMD sync.Map
 
 // hashToMD converts a hash.Hash implementation from this package to a GO_EVP_MD_PTR.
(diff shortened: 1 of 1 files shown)


simo5:
LGTM

simo5 (approved):

derekparker (approved):
LGTM

PR #77, opened 2023-06-15
Cache HMAC object and fix memory leak

qmuntal:
This PR improves `NewHMAC` so it reuses the same `EVP_MAC` object instead of fetching it every time.

Doing so also fixed a memory leak: the object returned by `EVP_MAC_fetch` should have been freed using `EVP_MAC_free`. The new approach doesn't require freeing that object, as it is kept as a global variable to be reused later on.

Thanks to caching, `NewHMAC` is now ~9% faster:

```go
goos: linux
goarch: amd64
pkg: github.com/golang-fips/openssl
cpu: AMD EPYC 7763 64-Core Processor                
           │   old.txt   │              new.txt               │
           │   sec/op    │   sec/op     vs base               │
NewHMAC-16   4.616µ ± 1%   4.199µ ± 1%  -9.04% (p=0.000 n=20)
```

### DIFF
--- a/hmac.go
+++ b/hmac.go
@@ -8,12 +8,15 @@ import "C"
 import (
 	"hash"
 	"runtime"
+	"sync"
 	"unsafe"
 )
 
+var paramDigest = C.CString("digest")
+
 var (
-	paramAlgHMAC = C.CString("HMAC")
-	paramDigest  = C.CString("digest")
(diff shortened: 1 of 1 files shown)


simo5 (approved):
LGTM

dagood (approved):

PR #86, opened 2023-06-21
Enforce using only non-deprecated functions

qmuntal:
This PR doesn't change functionality, but makes sure that all functions I've identified as deprecated in OpenSSL 3 are defined using the `DEFINEFUNC_LEGACY_1` macro, which will ensure we only use it in OpenSSL 1 code paths.

While here, I've removed a couple of unused functions.

Update: I've added the macro `DEFINEFUNC_LEGACY_1_1`, which can be used to define functions only available (and not deprecated) in OpenSSL 1.1.x. Thanks to that, we can now say that we don't use any deprecated function for any given OpenSSL version. Additionally, I've updated `checkheader` to validate that this remains true forever by using the `-DOPENSSL_NO_DEPRECATED` flag.

Update 2: Thanks top the previous check, I've found that `CRYPTO_set_id_callback` is deprecated in OpenSSL 1.0.2 We should be using [CRYPTO_THREADID_set_callback](https://www.openssl.org/docs/man1.0.2/man3/CRYPTO_THREADID_set_callback.html) instead.

### DIFF
--- a/cmd/checkheader/main.go
+++ b/cmd/checkheader/main.go
@@ -89,10 +89,10 @@ func gccRun(program string) error {
 	// any of the static checks fail. If it succeeds, it means
 	// the checked header matches the OpenSSL definitions.
 	p := exec.Command("gcc",
-		"-c",                           // skip linking
-		"-Werror",                      // promote all warnings to errors
-		"-Wno-deprecated-declarations", // deprecation warnings are expected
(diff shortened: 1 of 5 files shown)


dagood (approved):

dagood (approved):

qmuntal:
@derekparker @ueno please take a look at this PR.

ueno (approved):
LGTM, nice work!

Issue #87, opened 2023-06-27
Add API support for SHA3-384 and friends


xnox:
I was redirected here from https://github.com/microsoft/go-crypto-openssl/issues/57

In essence, I need golang toolchain that is FIPS compatible, and also is able to use crypto.SHA3_384 for which one normally needs import of `_ golang.org/x/crypto/sha3` that registers such a hash, given only sha3 stub without an implementation is part of toolchain's crypto module.

Given this is the easiest way to access and expose openssl APIs, will work on porting https://github.com/microsoft/go-crypto-openssl/pull/58 to this repo.

This would then open up easy access to FIPS-compliant sha3.

PR #88, opened 2023-06-28
sha: Add SHA3 implementation

xnox:
This adds OpenSSL SHA-3 family of hashes in the openssl package.

Separately sha3 package is added as a drop-in replacement for golang.org/x/crypto/sha3 package.

Together with FIPS-compliant Go toolchain one can thus attempt building FIPS-compliant applications that use "crypto" package and also need FIPS-compliant SHA-3.

Ideally, a patched golang toolchain will use openssl/xcrypto/sha3 in place of any imports of golang.org/x/crypto/sha3, this is a future improvement to ensure golang-fips toolchain provides FIPS SHA3 without source changes.

Fixes: https://github.com/golang-fips/openssl/issues/87

### DIFF
--- a/evp.go
+++ b/evp.go
@@ -31,6 +31,14 @@ func hashToMD(h hash.Hash) C.GO_EVP_MD_PTR {
 		ch = crypto.SHA384
 	case *sha512Hash:
 		ch = crypto.SHA512
+	case *sha3_224Hash:
+		ch = crypto.SHA3_224
+	case *sha3_256Hash:
+		ch = crypto.SHA3_256
+	case *sha3_384Hash:
+		ch = crypto.SHA3_384
+	case *sha3_512Hash:
+		ch = crypto.SHA3_512
 	}
 	if ch != 0 {
 		return cryptoHashToMD(ch)
@@ -72,6 +80,26 @@ func cryptoHashToMD(ch crypto.Hash) (md C.GO_EVP_MD_PTR) {
 		return C.go_openssl_EVP_sha384()
 	case crypto.SHA512:
 		return C.go_openssl_EVP_sha512()
+	case crypto.SHA3_224:
+		if !SupportsSHA3() {
+			return nil
+		}
+		return C.go_openssl_EVP_sha3_224()
+	case crypto.SHA3_256:
+		if !SupportsSHA3() {
+			return nil
+		}
+		return C.go_openssl_EVP_sha3_256()
+	case crypto.SHA3_384:
+		if !SupportsSHA3() {
+			return nil
+		}
+		return C.go_openssl_EVP_sha3_384()
+	case crypto.SHA3_512:
+		if !SupportsSHA3() {
+			return nil
+		}
+		return C.go_openssl_EVP_sha3_512()
 	}
 	return nil
 }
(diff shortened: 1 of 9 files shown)


evp.go
```diff
@@ -72,6 +80,26 @@ func cryptoHashToMD(ch crypto.Hash) (md C.GO_EVP_MD_PTR) {
 		return C.go_openssl_EVP_sha384()
 	case crypto.SHA512:
 		return C.go_openssl_EVP_sha512()
+	case crypto.SHA3_224:
+		if vMajor == 1 && (vMinor == 0 || vPatch == 0) {
```
ueno, line 84: It might be easier to read if it is written with `>=`, as in [`SuppertsHKDF`](https://github.com/golang-fips/openssl/blob/dc825f725ab84c5598da7463c302960c0283460d/hkdf.go#L16).

sha.go
```diff
@@ -559,6 +587,214 @@ func (h *sha512Hash) UnmarshalBinary(b []byte) error {
...
+	for i := 0; i < len(d.A); i++ {
+		b, d.A[i] = consumeUint64(b)
+	}
+	b, d.block_size = consumeUint64(b)
+	b, d.md_size = consumeUint64(b)
+	b, d.bufsz = consumeUint64(b)
+	b = b[copy(d.x[:], b):]
+}
+
+func (h *sha3_224Hash) MarshalBinary() ([]byte, error) {
```
qmuntal, line 702: The `x/crypto/sha3` package doesn't implement the marshaling functions. As doing so it quite tricky, I'll rather not do that. You can adapt the tests in `sha_test.go` to skip the checks that assume marshaling functions are implemented.

sha3/sha3.go
```diff
@@ -0,0 +1,27 @@
+package sha3
```
qmuntal, line 1: I would prefer to have the `sha3` package under a packaged named `xcrypto`, so it is clear that we are providing a replacement for `x/crypto` and not for the standard library.

sha3/sha3.go
```diff
@@ -0,0 +1,27 @@
...
+var (
+	New224 = openssl.NewSHA3_224
+	New256 = openssl.NewSHA3_256
+	New384 = openssl.NewSHA3_384
+	New512 = openssl.NewSHA3_512
+	Sum224 = openssl.SHA3_224
+	Sum256 = openssl.SHA3_256
+	Sum384 = openssl.SHA3_384
+	Sum512 = openssl.SHA3_512
+)
```
qmuntal, line 20: The Go compiler can't apply many important optimizations (e.g., inlining and escape analysis) when calling a function assigned to a `var`, as it's value can change at runtime.

Better define a wrapper functions, which won't have a performance impact as it will be inlined away.

```go
func New224() hash.Hash {
    return openssl.NewSHA3_224()
}
```

qmuntal (changes requested):

evp.go
xnox: I like SupportsHKDF function, if it's ok to export like that one, I think I should use that pattern for SHA3 as well. Let me rewrite those bits.

ueno (approved):

qmuntal (approved):

xnox:
please approve a workflow run, as this is my first contribution to golang-fips.

PR #89, opened 2023-07-10
Purge xcrypto/sha3 as it is unusable.

xnox:
Attempting to use this package directly only works with vanilla golang toolchain. A FIPS/boring golang toolchain, doesn't expose it's internal package which conflicts with this one at build time, thus it really is not possible to adapt application to use openssl-powered sha3 by doing module replace for this one.

Instead it is best to modify golang boring based toolchain to expose boring-only "crypto/sha3" public package, which directly uses the vendored copy, and then modify applications appropriately.

See examples at
https://code.launchpad.net/~xnox/go-snap/+git/go-snap/+merge/445900 and
https://github.com/snapcore/snapd/pull/12934/files#diff-7f06ad5451a2e1f2410e6c8b41eb402e15f14c52e1293a1f3bb7ec94b84f319d

Thus requesting to drop sha3 package from this internal-only to a given toolchain library.


qmuntal (approved):
Would have been nice. LGTM.

derekparker (approved):
LGTM

PR #90, opened 2023-07-11
remove !android build tag

qmuntal:
The `!android` build tag was inherited from the `boring` package, which doesn't support Android. OpenSSL does support building on Android, see [docs](https://github.com/openssl/openssl/blob/master/NOTES-ANDROID.md), so we shouldn't be artificially limiting the allowed environments.

This PR is more of a cleanup PR than a feature request, AFAIK no one is using it on Android devices yet.


ueno (approved):

derekparker (approved):
LGTM

PR #91, opened 2023-07-11
Bump minimum Go version to 1.20

qmuntal:
This module will be first used in go1.21, so we don't have to keep backwards compatibility with older versions.

Once we start using it, bumping the minimum Go version will be potentially more disruptive, better do it now.


dagood (approved):

derekparker (approved):
LGTM

PR #93, opened 2023-07-13
Fix HashVerifyECDSA function signature in notboring.go

dbenoit17:


openssl/notboring.go
```diff
@@ -61,7 +61,7 @@ func NewPublicKeyECDSA(curve string, X, Y BigInt) (*PublicKeyECDSA, error) {
 func HashSignECDSA(priv *PrivateKeyECDSA, hash []byte, h crypto.Hash) (r, s BigInt, err error) {
 	panic("boringcrypto: not available")
 }
-func HashVerifyECDSA(pub *PublicKeyECDSA, msg []byte, r, s *big.Int, h crypto.Hash) bool {
+func HashVerifyECDSA(pub *PublicKeyECDSA, msg []byte, r, s *BigInt, h crypto.Hash) bool {
```
ueno, line 64: I'm not sure if this alone is correct; in the current git master, `ecdsa.go` defines both `HashSignECDSA` and `HashVerifyECDSA` with `*big.Int`. Maybe they should also be using `*BigInt`?

karianna:
@dbenoit17 Are you able to address the comments in this review?

PR #94, opened 2023-07-31
Support multiple hkdf.Read calls

qmuntal:
This PR adds support for deriving multiple keys using the same hkdf instance. This is supported in `x/sys/hkdf`, but our implementation didn't and was instead returing the same derived key over and over. No API change is required.

Worth noting that calling `hkdf.Read` multiple times is much more slower than doing so with `x/sys/hkdf` because OpenSSL doesn't properly support partial incremental reads. Instead, we have to derive the whole key on every `hkdf.Read` call and just return the last unread section. Luckily, deriving many keys from the same hkdf instance is very uncommon, so I don't expect it to be a real problem.


hkdf.go
```diff
@@ -82,11 +84,20 @@ func (c *hkdf) finalize() {
...
-	outLen := C.size_t(len(p))
-	if C.go_openssl_EVP_PKEY_derive(c.ctx, base(p), &outLen) != 1 {
+	// EVP_PKEY_derive doesn't support incremental output, each call
+	// derives the key from scratch and returns the requested bytes.
+	// To implement io.Reader, we need to ask for len(c.buf) + len(p)
+	// bytes and copy the last derived len(p) bytes to p.
+	// We use c.buf to know how many bytes we've already derived and
+	// to avoid allocating the whole output buffer on each call.
+	start := len(c.buf)
+	c.buf = append(c.buf, make([]byte, len(p))...)
```
ueno, line 94: Maybe we should limit the total allocation to `255 * HashLen` as [defined](https://www.rfc-editor.org/rfc/rfc5869#section-2.3)? Although OpenSSL has the check in FIPS mode, that would prevent allocating unreasonable amount of memory prior to the EVP_PKEY_derive call.

hkdf.go
```diff
@@ -82,11 +84,20 @@ func (c *hkdf) finalize() {
...
+	outLen := C.size_t(len(c.buf))
+	if C.go_openssl_EVP_PKEY_derive(c.ctx, base(c.buf), &outLen) != 1 {
 		return 0, newOpenSSLError("EVP_PKEY_derive")
 	}
-	return int(outLen), nil
+	n := copy(p, c.buf[start:outLen])
```
ueno, line 99: Should we care about zeroization of the buffer content, in particular the previous output?

hkdf.go
qmuntal: Good point. OpenSSL fails if trying to read more than allowed, and we are validating that assumption with `TestHKDFLimit`. `x/crypto/hkdf` also fails in that case. If we limit the total allocations to `255*HashLen`, OpenSSL would not fail, which is not what we want. What I can to to avoid unnecessary allocations is to error out early if `len(p)+len(c.buf) > 255*hashLen`.

hkdf.go
ueno: Then that should be fine.

hkdf.go
ueno: I guess that would be sufficient for most use-cases, but if we want to strictly follow the io.Reader [convention](https://pkg.go.dev/io#Reader), shouldn't it return truncated output with EOF as err?

qmuntal: That's what I would have done too, but [x/crypto/hkdf#hkdf.Read](https://github.com/golang/crypto/blob/edc325d13aa98b5b79708f7122a33bf392cdc1a7/hkdf/hkdf.go#L46-L52) doesn't follow that convention, and we are trying to mimic what it does.

ueno: Perhaps this should be reported as a bug in upstream?

PR #97, opened 2023-08-01
Add !cmd_go_bootstrap tag

derekparker:
This is necessary for the Red Hat Go fork as our support for building with openssl is opt-out instead of opt-in, meaning it isn't gated behind a build tag. Without this change `./make.bash` will fail.


qmuntal (approved):
LGTM

PR #98, opened 2023-08-01
Support Unix and Windows

corhere:
- Carries #48 

Add support for any target which satisfies the build constraint `cgo && (windows || unix)`.

The Linux limitation is somewhat artificial. The only platform-specific parts of the module are dynamic linking and loading of the OpenSSL library, and the OpenSSL 1.0.2 threading callbacks. The implementations of both are already portable to any Unix-like OS as `dlfcn.h` and pthreads are part of the POSIX standard. The only issue with the dynamic loader implementation is that it assumes the naming convention for the library shared object is `"libcrypto.so.<version>"` which limits support for darwin and other platforms which follow a different convention. Change `openssl.Init()` and `openssl.CheckVersion()` to take the library file name as its argument to make the bindings agnostic to the target platform's library file naming convention. Widen the build constraints to allow the module with the `dlopen()`-based loader to be built on any OS which satisfies the "unix" Go build constraint.

Windows's dynamic library loader API is shaped very similarly to POSIX's, aside from the function names. Add support for Windows by adding Windows implementations for linking and loading the OpenSSL library, and the threading callbacks.

-----

Tested on Windows Server 2022 with OpenSSL 1.0.2zh-fips (Safelogic Cryptocomply), and macOS Ventura 13.4 (x86_64) with Homebrew OpenSSL 1.1.1 and OpenSSL 3.0.8.

### DIFF
--- a/ecdh.go
+++ b/ecdh.go
@@ -101,7 +101,7 @@ func (k *PrivateKeyECDH) PublicKey() (*PublicKeyECDH, error) {
 			return nil, newOpenSSLError("EVP_PKEY_get_octet_string_param")
 		}
 		bytes = C.GoBytes(unsafe.Pointer(cbytes), C.int(n))
-		C.free(unsafe.Pointer(cbytes))
+		cryptoFree(unsafe.Pointer(cbytes))
 	default:
 		panic(errUnsupportedVersion())
 	}
--- a/ecdh.go
+++ b/ecdh.go
@@ -314,8 +314,8 @@ func GenerateKeyECDH(curve string) (*PrivateKeyECDH, []byte, error) {
 	// generating a private ECDH key.
 	bits := C.go_openssl_EVP_PKEY_get_bits(pkey)
 	bytes := make([]byte, (bits+7)/8)
-	if C.go_openssl_BN_bn2binpad(priv, base(bytes), C.int(len(bytes))) == 0 {
-		return nil, nil, newOpenSSLError("BN_bn2binpad")
+	if err := bnToBinPad(priv, bytes); err != nil {
+		return nil, nil, err
 	}
 	k = &PrivateKeyECDH{pkey, curve, true}
 	runtime.SetFinalizer(k, (*PrivateKeyECDH).finalize)
--- a/ecdh.go
+++ b/ecdh.go
@@ -1,4 +1,4 @@
-//go:build linux && !cmd_go_bootstrap
+//go:build !cmd_go_bootstrap
 
 package openssl
 
--- a/init_unix.go
+++ b/init_unix.go
@@ -1,0 +1,31 @@
+//go:build unix && !cmd_go_bootstrap
+
+package openssl
+
+// #cgo LDFLAGS: -ldl
+// #include <stdlib.h>
+// #include <dlfcn.h>
+import "C"
+import (
+	"errors"
+	"unsafe"
+)
+
+func dlopen(file string) (handle unsafe.Pointer, err error) {
+	cv := C.CString(file)
+	defer C.free(unsafe.Pointer(cv))
+	handle = C.dlopen(cv, C.RTLD_LAZY|C.RTLD_LOCAL)
+	if handle == nil {
+		errstr := C.GoString(C.dlerror())
+		return nil, errors.New("openssl: can't load " + file + ": " + errstr)
+	}
+	return handle, nil
+}
+
+func dlclose(handle unsafe.Pointer) error {
+	if C.dlclose(handle) != 0 {
+		errstr := C.GoString(C.dlerror())
+		return errors.New("openssl: can't close libcrypto: " + errstr)
+	}
+	return nil
+}
--- a/init_windows.go
+++ b/init_windows.go
@@ -1,0 +1,36 @@
+//go:build !cmd_go_bootstrap
+
+package openssl
+
+import (
+	"syscall"
+	"unsafe"
+)
+
+type dlopenError struct {
+	file string
+	err  error
+}
+
+func (e *dlopenError) Error() string {
+	return "openssl: can't load " + e.file + ": " + e.err.Error()
+}
+
+func (e *dlopenError) Unwrap() error {
+	return e.err
+}
+
+func dlopen(file string) (handle unsafe.Pointer, err error) {
+	// As Windows generally does not ship with a system OpenSSL library, let
+	// alone a FIPS 140 certified one, use the default library search order so
+	// that we preferentially load the DLL bundled with the application.
+	h, err := syscall.LoadLibrary(file)
+	if err != nil {
+		return nil, &dlopenError{file: file, err: err}
+	}
+	return unsafe.Pointer(h), nil
+}
+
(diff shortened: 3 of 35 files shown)


ecdh.go
```diff
@@ -101,7 +99,7 @@ func (k *PrivateKeyECDH) PublicKey() (*PublicKeyECDH, error) {
 			return nil, newOpenSSLError("EVP_PKEY_get_octet_string_param")
 		}
 		bytes = C.GoBytes(unsafe.Pointer(cbytes), C.int(n))
-		C.free(unsafe.Pointer(cbytes))
```
qmuntal, line 104: Is this necessary for supporting Windows? If not, better do it in another PR.

openssl.go
```diff
@@ -216,6 +213,36 @@ func newOpenSSLError(msg string) error {
...
+// must be freed using C.go_openssl_CRYPTO_free. This function is equivalent to
+// the OPENSSL_malloc macro.
+//
+// Like C.malloc, this function is guaranteed to never return nil. If OpenSSL's
+// malloc indicates out of memory, it crashes the program.
+//
+// Only objects which the OpenSSL library will take ownership of (i.e. will be
+// freed by OPENSSL_free / CRYPTO_free) need to be allocated on the OpenSSL
+// heap.
+func cryptoMalloc(n int) unsafe.Pointer {
```
qmuntal, line 229: Is this necessary for supporting Windows? If not, better do it in another PR.

qmuntal (commented):
Thanks for working on this!

Please update [CI](https://github.com/golang-fips/openssl/blob/v2/.github/workflows/test.yml) to also run on Windows and Darwin test

ecdh.go
corhere: It's not _strictly_ necessary if the C toolchain used with cgo links against the same CRT as the OpenSSL DLL. I'll split this change into a separate PR.

corhere: The version of MinGW-w64 installed on the Windows Server 2022 GHA runner image links against a different CRT than the preinstalled OpenSSL libraries. Therefore this _is_ necessary to support running Windows CI tests on GHA using the default toolchains.

corhere:
Apologies for the force-push spam; it took me a few tries to get the GitHub Actions workflow in a good state. (I never did discover where on the macOS runner image libcrypto.1.1.dylib is hiding...) This PR should be ready for review now.

init_unix.go
```diff
@@ -0,0 +1,31 @@
+//go:build unix
```
dagood, line 1:
```suggestion
//go:build unix && !cmd_go_bootstrap
```

thread_setup_windows.c
```diff
@@ -0,0 +1,31 @@
+#include "goopenssl.h"
```
dagood, line 1: Condition for consistency's sake?

```suggestion
//go:build windows

#include "goopenssl.h"
```

init_windows.go
```diff
@@ -0,0 +1,22 @@
+package openssl
```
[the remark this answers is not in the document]
corhere: Good call. While this file doesn't `import "C"` currently, that could change at any time.

openssl.go
```diff
@@ -34,25 +33,25 @@ var nativeEndian binary.ByteOrder
...
+// Subsequent calls will return the same error result as the one from the first call.
 //
-// version will be appended to the OpenSSL shared library name as a version suffix
-// when calling dlopen. For example, `version=1.1.1k-fips` makes Init look for
-// the shared library libcrypto.so.1.1.1k-fips.
-func Init(version string) error {
+// The file is passed to dlopen() verbatim to load the OpenSSL shared library.
+// For example, `file=libcrypto.so.1.1.1k-fips` makes Init look for the shared
+// library libcrypto.so.1.1.1k-fips.
+func Init(file string) error {
```
derekparker, line 54: This change is not backwards compatible and will break existing users of the `openssl/v2` module.

How would everyone feel about leaving this API as-is and updating it to prepend `libcrypto.so` like `dlopen` used to, and exporting a new API like `InitFullPath` or similar? A change like this should technically wait for `v3`.

derekparker (changes requested):
One concern, otherwise would like @ueno eyes on this.

openssl.go
qmuntal: I'm fine with changing the meaning of the `Init` parameter. Technically, `v2` hasn't been officially released yet, it is still in release candidate phase (https://github.com/golang-fips/openssl/tags), so this would be an allowed breaking change.

The Microsoft fork started to use `openssl/v2` in the current Go development cycle (go1.22), so this wouldn't affect our release branches.

Having said this, I don't have an strong opinion here. I'm also good with having something like `InitPath`.

derekparker: I suppose you're right, and we're still integrating it on our end and haven't shipped anything with it yet so not a huge deal to update our code.

derekparker (approved):

init_windows.go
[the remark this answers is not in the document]
dagood: Ah, I didn't catch that `import "C"` was actually the reason to have this condition, thanks for the extra info there. I only saw a difference that I didn't understand. 😄

dagood (approved):

init_windows.go
```diff
@@ -0,0 +1,36 @@
...
+	return e.err
+}
+
+func dlopen(file string) (handle unsafe.Pointer, err error) {
+	// As Windows generally does not ship with a system OpenSSL library, let
+	// alone a FIPS 140 certified one, use the default library search order so
+	// that we preferentially load the DLL bundled with the application.
+	h, err := syscall.LoadLibrary(file)
+	if err != nil {
+		return nil, &dlopenError{file: file, err: err}
```
qmuntal, line 29: Why returning a typed error instead of just `errors.New("openssl: can't load " + e.file + ": " + err.Error())`

ueno (approved):

qmuntal (changes requested):
Some files have been added since this PR was posted, i.e. https://github.com/golang-fips/openssl/blob/v2/tls1prf.go and https://github.com/golang-fips/openssl/blob/v2/tls1prf_test.go. Please update their build constraints too.

init_windows.go
corhere: So the caller (of `openssl.Init()`) is able to interrogate the wrapped `syscall.Errno` error returned from `LoadLibrary` for the cause of the failure, e.g. to differentiate `errors.Is(err, syscall.ERROR_ACCESS_DENIED)` from `errors.Is(err, syscall.ERROR_MOD_NOT_FOUND)`.

In contrast, POSIX `dlopen()` only surfaces errors through the `dlerror()` API, which returns a human readable string rather than an enumerated error. The caller would not be able to do much programmatically with that information aside from surfacing it up to the user verbatim, so nothing of value is lost by returning a plain `errors.New()` string there.

qmuntal: Thanks, makes sense.

qmuntal (approved):

qmuntal:
Merging. Thanks!

PR #101, opened 2023-08-08
Generalize checking if a crypto.Hash is supported

qmuntal:
This PR removes `SupportsSHA3()` and replaces it with `SupportsHash(crypto.Hash)`, which is more future-proof and supports fine-grained hash queries.

### DIFF
--- a/sha.go
+++ b/sha.go
@@ -63,7 +63,7 @@ func SHA512(p []byte) (sum [64]byte) {
 	return
 }
 
-// SupportsHash returns true if h is supported.
+// SupportsHash returns true if a hash.Hash implementation is supported for h.
 func SupportsHash(h crypto.Hash) bool {
 	switch h {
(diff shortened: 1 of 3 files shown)


sha.go
```diff
@@ -63,11 +63,17 @@ func SHA512(p []byte) (sum [64]byte) {
 	return
 }
 
-// Same as SupportsHKDF, as in v1.1.1+
-func SupportsSHA3() bool {
-	return vMajor > 1 ||
-		(vMajor >= 1 && vMinor > 1) ||
-		(vMajor >= 1 && vMinor >= 1 && vPatch >= 1)
+// SupportsHash returns true if h is supported.
```
dagood, line 66: Given the comment above about RSA and MD5, this seems too general and I wonder what "supported" means. Just a little extra detail seems like enough to me:

```suggestion
// SupportsHash returns true if a hash.Hash implementation is supported for h.
```

dagood (approved):
Other than the doc comment, LGTM.

PR #102, opened 2023-08-10
Support PBKDF2

qmuntal:
All OpenSSL versions supports [PKCS5_PBKDF2_HMAC](https://www.openssl.org/docs/manmaster/man3/PKCS5_PBKDF2_HMAC.html) and, fortunately, it has the same parameters as [x/crypto/pbkdf2.Key](https://pkg.go.dev/golang.org/x/crypto/pbkdf2#Key).


ueno (approved):
Looks good to me.

PR #103, opened 2023-08-15
Support TLS 1.2 PRF

ueno:
Go 1.21 implements TLS extended master secret (RFC 7627), which uses a different label ("extended master secret") than the original TLS 1.2 PRF ("master secret").  The use of EMS is now required by FIPS 140-3 IG D.Q., and thus the logic needs to be provided by a FIPS certified module (OpenSSL).  This adds a new function openssl.TLS1PRF for that.


qmuntal (approved):
LGTM!

derekparker (approved):
lgtm

ueno:
@derekparker rebased

@qmuntal note that this doesn't support empty "secret" nor "salt", because the underlying `OSSL_PARAM_construct_octet_string` doesn't accept (NULL, 0) parameter. I guess that shouldn't be a problem as those never be empty in the TLS 1.2 use-case.

qmuntal:
> @qmuntal note that this doesn't support empty "secret" nor "salt", because the underlying OSSL_PARAM_construct_octet_string doesn't accept (NULL, 0) parameter. I guess that shouldn't be a problem as those never be empty in the TLS 1.2 use-case.

No problem here. It doesn't make sense to generate a PRF without secret nor salt, and this behavior is documented in the OpenSSL docs:

> The digest, secret value and seed must be set before a key is derived or an error occurs.
https://www.openssl.org/docs/manmaster/man3/EVP_PKEY_CTX_set1_tls1_prf_secret.html

PR #108, opened 2023-08-23
Support crypto/des

qmuntal:
This PR adds support for the ciphers defined in [crypto/des](https://pkg.go.dev/crypto/des). Note that DES doesn't support CTR nor GCM modes, that's by design.

This is the new public API:

```go
// SupportsDESCipher returns true if NewDESCipher is supported.
func SupportsDESCipher() bool

// SupportsTripleDESCipher returns true if NewTripleDESCipher is supported.
func SupportsTripleDESCipher() bool

func NewDESCipher(key []byte) (cipher.Block, error)

func NewTripleDESCipher(key []byte) (cipher.Block, error)
```

The PR contains a lot of changes because I've refactored AES so it can share almost all the code with DES.

Please review the PR commit by commit, it will be easier. The commit with more changes is 8cf7f7019fce8a9dcd38ddf363cb4383998d544a, in which I moved most of the generalized AES code into `cipher.go`. The commit ed657596b31e1fbeb13c5a445b6f17d7ca0a6a6b is not required, as we know the block size for the ciphers we support, but IMO it is more robust to take them from OpenSSL.


shims.h
```diff
@@ -9,12 +9,6 @@ enum {
     GO_OPENSSL_INIT_LOAD_CONFIG = 0x00000040L
 };
 
-// #include <openssl/aes.h>
-enum {
-    GO_AES_ENCRYPT = 1,
-    GO_AES_DECRYPT = 0
-};
```
dagood, line 16: I think it makes sense to get rid of these particular constants (they aren't accurate given our use and they aren't mentioned in `EVP_CipherInit_ex` docs), but... I kind of wish we still had some constants of our own rather than putting 0 and 1 in the C and Go code. 😅 (And in theory, -1 is also a possibility.)

des.go
```diff
@@ -0,0 +1,82 @@
...
+	"runtime"
+)
+
+const desBlockSize = 8
+
+// SupportsDESCipher returns true if NewDESCipher is supported.
+func SupportsDESCipher() bool {
+	// True for stock OpenSSL 1.
+	// False for stock OpenSSL 3 unless the legacy provider is available.
+	return loadCipher(cipherDES, cipherModeECB) != nil
```
dagood, line 16: It makes sense to me that `cipherModeECB` would always be the same as `cipherModeCBC`, but maybe worth pointing out in the comment so this doesn't look like it might be making a bad assumption?

(Also in `SupportsTripleDESCipher`)

aes_test.go
```diff
@@ -1,27 +1,35 @@
...
-	gi, err := c.NewGCM(gcmStandardNonceSize, gcmTagSize)
+	const (
+		gcmTagSize           = 16
+		gcmStandardNonceSize = 12
+	)
+
+	c := ci.(interface {
+		NewGCM(nonceSize, tagSize int) (cipher.AEAD, error)
+	})
+	g, err := c.NewGCM(gcmStandardNonceSize, gcmTagSize)
```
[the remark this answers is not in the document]
qmuntal: The type is an implementation details that I would prefer not to export. The idea is to test the `openssl` package as an external consumer, which won't have access to `aesCipher` but would have to type-cast the interface.

About the constants, it would be helpful to export them in `export_test.go`, but the idiom is to just export what's indispensable for testing (ideally nothing). If something can be recreated from outside the tested package, as an external consumer would, then better not export it.

shims.h
qmuntal: Good point, I'll add that constant. FYI, -1 is documented to be a valid value:

> The operation performed depends on the value of the enc parameter. It should be set to 1 for encryption, 0 for decryption and -1 to leave the value unchanged (the actual value of 'enc' being supplied in a previous call).
https://www.openssl.org/docs/man3.0/man3/EVP_CipherInit_ex.html

qmuntal:
@ueno @derekparker Please take a look

qmuntal:
@ueno @derekparker Please take a look

ueno (approved):
Looks good to me.

PR #109, opened 2023-08-24
Add `label` parameter to TLS1PRF

qmuntal:
This PR updates the `TLS1PRF` signature to accept the PRF label as a separate parameter:

```go
func TLS1PRF(secret, label, seed []byte, keyLen int, h func() hash.Hash) ([]byte, error)`
```

It is common for TLS PRF functions to accept the `label` as a separate parameter instead of prepending it to the `seed`. This better match the `crypto/tls` PRF signature: https://github.com/golang/go/blob/d5c5808534f0ad97333b1fd5fff81998f44986fe/src/crypto/tls/prf.go#L48


dagood (approved):
My only thought is: maybe the performance will be different if we append the seed on the Go side rather than call into cgo twice to have OpenSSL do it.

ueno (approved):
+1

qmuntal:
> My only thought is: maybe the performance will be different if we append the seed on the Go side rather than call into cgo twice to have OpenSSL do it.

If that slowdown is relevant, then we can always implement part of the function in C while still having a better API 😸

PR #111, opened 2023-08-24
Fix AES-GCM decryption on OpenSSL 1.0.2-fips

corhere:
Some of the standard library crypto/cipher AES-GCM tests fail when using OpenSSL 1.0.2 in FIPS mode, but pass with the same library when FIPS mode is disabled. Extend the AES-GCM tests to cover all the failure cases caught by the standard library tests, build OpenSSL 1.0.2 in CI to be FIPS-enabled, and update the bindings so the tests pass.


qmuntal (approved):
LGTM. Thanks!

ueno (approved):
LGTM!

qmuntal (changes requested):
Commit 7bb4f4b6dd147ab34a749eb024995a629c00b175 should limit its changes to OpenSSL 1.0.2, other versions shouldn't need the `FIPS()` call (as per the commit description) to know if an algorithm is supported or not.

PR #114, opened 2023-09-01
Support Ed25519

qmuntal:
This PR adds support for Ed25519.

OpenSSL still doesn't support Ed25519ctx nor Ed25519ph, so these are not added.

New APIs:

```go
func SupportsEd25519() bool
func GenerateKeyEd25519() (pub, priv []byte, err error)
func NewKeyFromSeedEd25519(seed []byte) (priv []byte, err error)
func SignEd25519(priv, message []byte) (sig []byte, err error)
func VerifyEd25519(pub, message, sig []byte)
```


ed25519.go
```diff
@@ -0,0 +1,170 @@
...
+	err = extractPKEYPrivEd25519(pkey, priv)
+	if err != nil {
+		return nil, nil, err
+	}
+	pub = make([]byte, publicKeySizeEd25519)
+	copy(pub, priv[seedSizeEd25519:])
+	return pub, priv, nil
+}
+
+func NewKeyFromSeedEd25519(seed []byte) (priv []byte, err error) {
```
dagood, line 69: I think it might be worth mentioning that "seed" is based on the naming that crypto/ed25519 uses, and that's also why the private key includes the public key inside it. It might be obvious enough to check that given the purpose of golang-fips, but it took me a bit because I was looking at the OpenSSL side first.

(Maybe simply copy the package doc from crypto/ed25519 to this file for reference?)

ed25519.go
```diff
@@ -0,0 +1,170 @@
...
+	err = signEd25519(sig, priv, message)
+	if err != nil {
+		return nil, err
+	}
+	return sig, err
+}
+
+func signEd25519(sig, priv, message []byte) error {
+	if l := len(priv); l != privateKeySizeEd25519 {
+		panic("ed25519: bad private key length: " + strconv.Itoa(l))
```
dagood, line 120: Nit: Not a fan of `l` one-letter variables when `1` is also around--but at least it's in a very small scope so it's probably fine.

ed25519.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	if siglen != signatureSizeEd25519 {
+		panic("ed25519: bad signature length: " + strconv.Itoa(int(siglen)))
+	}
+	return nil
+}
+
+func VerifyEd25519(pub, message, sig []byte) error {
+	return verifyEd25519(pub, message, sig)
+}
```
dagood, line 147: Why is this one a wrapper even though nothing's being outlined?

qmuntal: It's a left-over, I'll fix that.

ed25519.go
```diff
@@ -0,0 +1,170 @@
...
+	// privateKeySize is the size, in bytes, of private keys as used in crypto/ed25519.
+	privateKeySizeEd25519 = 64
+	// signatureSize is the size, in bytes, of signatures generated and verified by crypto/ed25519.
+	signatureSizeEd25519 = 64
+	// seedSize is the size, in bytes, of private key seeds. These are the private key representations used by RFC 8032.
+	seedSizeEd25519 = 32
+)
+
+// TODO: Add support for Ed25519ph and Ed25519ctx when OpenSSL supports them,
+// which will probably be in 3.2.0.
```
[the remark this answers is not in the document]
qmuntal: Done!

dagood (approved):

qmuntal:
@ueno @derekparker PTAL

qmuntal:
@derekparker @ueno I'll be taking a long leave soon and I'll like to have this PR as ready as possible to be merged. Could you take a look? Thanks!

ed25519.go
```diff
@@ -0,0 +1,173 @@
...
+		return newOpenSSLError("EVP_PKEY_get_raw_private_key")
+	}
+	if privSize != seedSizeEd25519 {
+		panic("ed25519: bad private key length: " + strconv.Itoa(int(privSize)))
+	}
+	return nil
+}
+
+// SignEd25519 signs the message with priv and returns a signature.
+func SignEd25519(priv, message []byte) (sig []byte, err error) {
```
ueno, line 114: Unlike ECDSA, this API takes a private key as bytes and internally recovers the corresponding EVP_PKEY; that I suspect would prevent using non-extractable keys such as on PKCS#11. Why not defining a dedicated key types around EVP_PKEY?

qmuntal: Good point. I've updated the implementation to use `PrivateKeyEd25519` and `PublicKeyEd25519` instead of raw byte slices.

ueno (approved):
Looks good to me; thank you for the update!

PR #116, opened 2023-09-08
tls1prf: require callers to pass in the result buffer

qmuntal:
This PR updates `TLS1PRF` to accept a byte slice parameter where to write the output. This avoids allocating a new slice on each function call and integrates better with the standard library, which expects the PRF to update an already existing slice: https://github.com/golang/go/blob/2f0b28da1900909a2c3ddf646bb508fc7effb8f2/src/crypto/tls/prf.go#L68.

To make is clear, the current code would have to be integrated like this:

```go
func prf12(hashFunc func() hash.Hash) func(result, secret, label, seed []byte) error {
	return func(result, secret, label, seed []byte) error {
		if backend.Enabled && backend.SupportsTLS1PRF() {
			out, err := backend.TLS1PRF(secret, label, seed, len(result), hashFunc)
			if err != nil {
				return fmt.Errorf("crypto/tls: prf12: %v", err)
			}
			copy(result, out)
			return nil
		}
                ...
	}
}
```

While with the new approach, it would like this:

```go
func prf12(hashFunc func() hash.Hash) func(result, secret, label, seed []byte) error {
	return func(result, secret, label, seed []byte) error {
		if backend.Enabled && backend.SupportsTLS1PRF() {
			err := backend.TLS1PRF(result, secret, label, seed, len(result), hashFunc)
			if err != nil {
				return fmt.Errorf("crypto/tls: prf12: %v", err)
			}
			return nil
		}
                ...
	}
}
```

### DIFF
--- a/tls1prf.go
+++ b/tls1prf.go
@@ -16,7 +16,10 @@ func SupportsTLS1PRF() bool {
 		(vMajor >= 1 && vMinor >= 1)
 }
 
-func TLS1PRF(secret, label, seed []byte, keyLen int, h func() hash.Hash) ([]byte, error) {
+// TLS1PRF implements the TLS 1.0/1.1 pseudo-random function if h is nil or crypto.MD5SHA1,
+// else it implements the TLS 1.2 pseudo-random function.
+// The pseudo-random number will be written to result and will be of length len(result).
+func TLS1PRF(result, secret, label, seed []byte, h func() hash.Hash) error {
 	var md C.GO_EVP_MD_PTR
 	if h == nil {
 		// TLS 1.0/1.1 PRF doesn't allow to specify the hash function,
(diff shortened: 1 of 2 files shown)


ueno (approved):
Makes sense to me; I thought it might be good to align with other KDF API (e.g., either `crypto/pbkdf2` or `crypto/hkdf`), though TLS1PRF is purely for internal use, so it would be better to align with the actual usage in `crypto/tls`.

tls1prf.go
```diff
@@ -29,70 +32,71 @@ func TLS1PRF(secret, label, seed []byte, keyLen int, h func() hash.Hash) ([]byte
...
-		return nil, newOpenSSLError("EVP_PKEY_derive")
+	outLen := C.size_t(len(result))
+	if C.go_openssl_EVP_PKEY_derive(ctx, base(result), &outLen) != 1 {
+		return newOpenSSLError("EVP_PKEY_derive")
 	}
-	if outLen != C.size_t(keyLen) {
-		return nil, errors.New("tls1-prf: entropy limit reached")
+	// The Go standard library expects TLS1PRF to return the requested number of bytes,
+	// fail if it doesn't.
+	if outLen != C.size_t(len(result)) {
```
dagood, line 98: We ended up with a more detailed comment on the CNG side to clarify this is just a sanity check (rather than something the caller has to be specifically careful to avoid), maybe include something similar here?

```suggestion
	// The Go standard library expects TLS1PRF to return the requested number of bytes,
	// fail if it doesn't. While there is no known situation where this will happen,
	// EVP_PKEY_derive handles multiple algorithms and there could be a subtle mismatch
	// after more code changes in the future.
	if outLen != C.size_t(len(result)) {
```

dagood (approved):

PR #117, opened 2023-09-14
Add support for RC4

qmuntal:
This PR implements the necessary functions to support [crypto/rc4](https://pkg.go.dev/crypto/rc4).

New benchmark results:

```
goos: windows
goarch: amd64
BenchmarkRC4_128-12      5396026               220.1 ns/op       581.51 MB/s           0 B/op          0 allocs/op
BenchmarkRC4_1K-12        704779              1532 ns/op         668.28 MB/s           0 B/op          0 allocs/op
BenchmarkRC4_8K-12         96313             12660 ns/op         639.48 MB/s           0 B/op          0 allocs/op
```


cipher.go
```diff
@@ -477,17 +483,33 @@ func sliceForAppend(in []byte, n int) (head, tail []byte) {
...
+		// RC4 cipher supports a variable key length.
+		// We need to set the key length before setting the key,
+		// and to do so we need to have an initialized cipher ctx.
+		if C.go_openssl_EVP_CipherInit_ex(ctx, cipher, nil, nil, nil, C.int(encrypt)) != 1 {
+			return nil, newOpenSSLError("EVP_CipherInit_ex")
+		}
+		if C.go_openssl_EVP_CIPHER_CTX_set_key_length(ctx, C.int(len(key))) != 1 {
+			return nil, newOpenSSLError("EVP_CIPHER_CTX_set_key_length")
+		}
+		cipher = nil // don't reset the cipher
```
dagood, line 510: Setting `cipher` to `nil` in order to *not* reset `cipher` confused me to begin with. 😄 It took me a bit to figure out this was referring to this usage pattern mentioned in the OpenSSL docs:

> It is possible to set all parameters to NULL except type in an initial call and supply the remaining parameters in subsequent calls, all of which have type set to NULL. This is done when the default cipher parameters are not appropriate.

I think this would have connected it better for me:

```suggestion
		// Pass nil to the next call to EVP_CipherInit_ex to avoid resetting ctx's cipher.
		cipher = nil
```

dagood (approved):

qmuntal:
@derekparker @ueno I'll be taking a long leave soon and I'll like to have this PR as ready as possible to be merged. Could you take a look? Thanks!

ueno (approved):
LGTM!

PR #120, opened 2023-09-20
Read signed message through io.Reader in HashSign/HashVerify

ueno:
When signed message is large, passing it through a byte array would cost memory; this switches to using io.Reader to avoid that.

Using `io.Reader` in the function signatures could also help diversify those HashSign/HashVerify from Sign/Verify, through a separate interface, alongside the `crypto.Signer` interface.

### DIFF
--- a/ecdsa.go
+++ b/ecdsa.go
@@ -7,6 +7,7 @@ import "C"
 import (
 	"crypto"
 	"errors"
+	"io"
 	"runtime"
 )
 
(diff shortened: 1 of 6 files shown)


ueno:
The idea behind this is to allow one-shot signature interface to be defined alongside `crypto.Signer`, something like:

```go
type HashSigner interface {
  HashSign(rand, message io.Reader, opts SignerOpts) (signature []byte, err error)
}
```

and can be used like:
```go
if hs, ok := priv.(crypto.HashSigner); ok {
  sig, err = hs.HashSign(config.rand(), bytes.NewReader(msg), signOpts)
  ...
}
sig, err := priv.Sign(config.rand(), signed, signOpts)
...
```
as suggested in https://go.dev/blog/module-compatibility

qmuntal (approved):
LGTM!

ueno:
> The idea behind this is to allow one-shot signature interface to be defined alongside crypto.Signer, something like:

FYI, here is a [proposal](https://github.com/golang/go/issues/63405) for this; I'll try to see how well the HashSign/HashVerify could align with it.

derekparker (approved):
LGTM

karianna:
@ueno - Assuming this is still valid, are you able to rebase?

PR #121, opened 2023-09-21
Reconcile AES and DES error messages with upstream

qmuntal:
`crypto/aes` and `crypto/des` expects that `Encrypt` and `Decrypt` error messages start with `crypto/aes: ` and `crypto/des`, there is even a test for that: https://github.com/golang/go/blob/ace1494d9235be94f1325ab6e45105a446b3224c/src/crypto/aes/aes_test.go#L322-L333

We recently changed those methods to return the more generic prefix `crypto/cipher: `, which broke the upstream test.

This PR fixes that by using the appropriate error prefix.

### DIFF
--- a/aes.go
+++ b/aes.go
@@ -56,11 +56,17 @@ func (c *aesCipher) BlockSize() int {
 }
 
 func (c *aesCipher) Encrypt(dst, src []byte) {
-	c.encrypt(dst, src)
+	if err := c.encrypt(dst, src); err != nil {
+		// Upstream expects that the panic message starts with "crypto/aes: ".
+		panic("crypto/aes: " + err.Error())
+	}
 }
 
 func (c *aesCipher) Decrypt(dst, src []byte) {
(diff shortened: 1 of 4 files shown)


dagood (approved):

aes.go
```diff
@@ -56,11 +56,17 @@ func (c *aesCipher) BlockSize() int {
 }
 
 func (c *aesCipher) Encrypt(dst, src []byte) {
-	c.encrypt(dst, src)
+	if err := c.encrypt(dst, src); err != nil {
+		// Upstream expects that the panic message starts with "crypto/aes: ".
```
dagood, line 60: Hmm, I suppose "upstream" is not necessarily clear inside a technically independent module like this.

```suggestion
		// The Go crypto packages expect that the panic message starts with "crypto/aes: ".
```

qmuntal: Good point.

PR #122, opened 2023-09-28
Free thread-local OpenSSL state on thread exit

corhere:
OpenSSL 1.0.x automatically allocates a per-thread error queue as needed, but requires that the application call ERR_remove_thread_state() when terminating a thread to prevent the memory from leaking. As the Go runtime may create and terminate threads arbitrarily, the amount of memory leaked has the potential to grow unbounded.

Clean up the OpenSSL error queue for a thread upon thread exit. Use the native threading APIs to arrange a callback on thread exit as the Go runtime does not provide any similar facilities itself.

No explicit thread-exit handling is needed for OpenSSL 1.1.0 and above as it automatically deallocates thread-local resources without application assistance. See the OPENSSL_thread_stop(3) man page for more information.

### DIFF
--- a/goopenssl.h
+++ b/goopenssl.h
@@ -4,6 +4,8 @@
 
 #include "shims.h"
 
+// Suppress warnings about unused parameters.
+#define UNUSED(x) (void)(x)
 
 static inline void
 go_openssl_do_leak_check(void)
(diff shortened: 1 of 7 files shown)


derekparker:
cc @ueno

ueno (approved):
Thank you for the patch. It looks good to me.

qmuntal (approved):

PR #124, opened 2023-10-13
Update GenerateKeyEd25519 to only generate a private key

qmuntal:
While integrating Ed25519 into the standard library I noticed that it is not necessary for `GenerateKeyEd25519` to return the public key, as it can already be derived from the private key.

Doing so can help reduces allocations and cgo calls in [ed25519.GenerateKey](https://pkg.go.dev/crypto/ed25519#GenerateKey), depending on how one decides to implement it.

```cmd
goos: windows
goarch: amd64
                      │    sec/op    │   sec/op     vs base                │
Ed25519GenerateKey-12   47.48µ ± 18%   35.97µ ± 7%  -24.25% (p=0.000 n=10)

                      │  old.txt   │              new.txt               │
                      │    B/op    │    B/op     vs base                │
Ed25519GenerateKey-12   56.00 ± 0%   16.00 ± 0%  -71.43% (p=0.000 n=10)

                      │  old.txt   │              new.txt               │
                      │ allocs/op  │ allocs/op   vs base                │
Ed25519GenerateKey-12   4.000 ± 0%   2.000 ± 0%  -50.00% (p=0.000 n=10)
```

### DIFF
--- a/ed25519.go
+++ b/ed25519.go
@@ -85,25 +85,27 @@ func (k *PrivateKeyEd25519) Bytes() ([]byte, error) {
 	return priv, nil
 }
 
-// GenerateKeyEd25519 generates a public/private key pair.
-func GenerateKeyEd25519() (*PublicKeyEd25519, *PrivateKeyEd25519, error) {
-	pkeyPriv, err := generateEVPPKey(C.GO_EVP_PKEY_ED25519, 0, "")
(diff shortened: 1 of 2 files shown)


ueno (approved):
Looks like a significant speed-up; LGTM!

PR #125, opened 2023-10-16
rsa: skip setting label for RSA OAEP if empty

ueno:
The label parameter for RSA OAEP is optional and by default it should be empty.  Given that it is set only once in setupEVP, we can bypass the call to EVP_PKEY_CTX_set0_rsa_oaep_label if the label is empty to spare 1-byte memory allocation and the following CGO call.


qmuntal (approved):
Good catch. LGTM!

derekparker (approved):
lgtm

PR #126, opened 2023-10-17
v1: Port HMAC to using EVP_DigestSign if available

ueno:
This makes the HMAC binding use the EVP_DigestSign interface instead of the deprecated HMAC_CTX functions, so that the implementation go through the FIPS provider when used with OpenSSL 3.


ueno:
@dbenoit17 @derekparker this is the last (cryptographic) part the v1 binding uses a legacy API. Could you check when you have time?

derekparker (approved):
LGTM

Issue #128, opened 2023-11-04
Remove #cgo noescape nocallback directives


qmuntal:
Thanks to the upstream issue https://github.com/golang/go/issues/63293,  reported by @corhere, the Go team has decided to revert some buggy #cgo directives used in this project. We need to stop using them, as they won't compile anymore on tip.

Plain removal is fine, as they are just an optimization to avoid some allocations. Creating a C wrapper that don't allocate is even better.

@dagood @gdams @corhere

corhere:
We don't _necessarily_ need to stop using them until Go 1.23. The bug is that pre-go1.22 versions of `go mod vendor` fall over when encountering the offending directives, with no workaround. For most modules that would be a serious issue since any (transitive) importer of the module building their project using either in-support version of Go could be affected. But this is not most modules. The primary consumers are forks of the Go toolchain, so there are no transitive importers of the module. And aside from the private toolchain fork I am maintaining, all consumers are forks of Go 1.22+ which (until it was disabled) understand the directives. We're all necessarily patching the toolchain already to incorporate the OpenSSL bindings, so adding one more patch to the toolchain to re-enable the directives, either by reverting https://github.com/golang/go/commit/607e020150f1bba84beec11aceadbc42abf33986 or by special-casing this module so the directives are still disabled for application code, would not add much to the ongoing maintenance cost of our forks.

However:
- Someone would have to write and test the special-casing patch if reenabling the directives for application code is unacceptable
- Our forked toolchains and the applications built with them would be the only ones exercising the new cgo directives on Go 1.22, so we'd be on our own if the codegen is buggy
- Applications would not be able to import golang-fips/openssl/v2 until Go 1.23, though it is [unlikely](https://pkg.go.dev/github.com/golang-fips/openssl/v2?tab=importedby) to be an [issue](https://pkg.go.dev/github.com/golang-fips/openssl@v0.0.0-20230605154532-724e32b0f4b8/openssl?tab=importedby) in [practice](https://pkg.go.dev/github.com/microsoft/go-crypto-openssl@v0.2.8/openssl?tab=importedby)
- Patched Go 1.22 toolchains would be required for development and testing the OpenSSL bindings until Go 1.22 is released in February and the new cgo directives are reenabled on gotip
- It's still a nontrivial amount of work, and risk, all just for some micro-optimizations which will be generally available in Go 1.23

On the balance, I agree that removing the directives is the right call.

dagood:
[...]

PR #113, opened 2023-08-30
Add noescape and nocallback directive to EVP_PKEY_derive

qmuntal:
`noescape` and `nocallback` are new go1.22 cgo directives that helps the Go compiler take better decisions. See https://github.com/golang/go/issues/56378 for more context.

In this PR I've tagged [EVP_PKEY_derive](https://www.openssl.org/docs/manmaster/man3/EVP_PKEY_derive.html) with both directives, which allows the `keylen` parameter to stay on the stack. `EVP_PKEY_derive` is currently used in ECDH, HKDF and TLS1PRF, so they will all automatically benefit from this optimization.

Benchmark prove for ECDH:

```
goos: windows
goarch: amd64
pkg: github.com/golang-fips/openssl/v2
cpu: Intel(R) Core(TM) i7-10850H CPU @ 2.70GHz
        │   old.txt    │            new.txt             │
        │    sec/op    │    sec/op     vs base          │
ECDH-12   53.01µ ± 12%   53.24µ ± 11%  ~ (p=0.853 n=10)

        │  old.txt   │              new.txt               │
        │    B/op    │    B/op     vs base                │
ECDH-12   40.00 ± 0%   32.00 ± 0%  -20.00% (p=0.000 n=10)

        │  old.txt   │              new.txt               │
        │ allocs/op  │ allocs/op   vs base                │
ECDH-12   2.000 ± 0%   1.000 ± 0%  -50.00% (p=0.000 n=10)
```

PR #129, opened 2023-11-06
Remove noescape, nocallback cgo directives: reverted from go1.22

dagood:
* Fixes https://github.com/golang-fips/openssl/issues/128
* Removes file added in https://github.com/golang-fips/openssl/pull/113


derekparker (approved):
lgtm

PR #131, opened 2023-11-08
Wrap multi-return APIs using structs: avoid heap escape

dagood:
* For https://github.com/golang-fips/openssl/issues/128
* Replaces the optimization removed in https://github.com/golang-fips/openssl/pull/129
* I prefer this approach over https://github.com/golang-fips/openssl/pull/130, because this approach more faithfully passes data to/from the OpenSSL API to deal with in Go code.

```
goos: linux
goarch: amd64
cpu: Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz
       │   old.txt   │        new-struct.txt        │
       │   sec/op    │   sec/op     vs base         │
ECDH-8   104.1µ ± 3%   103.8µ ± 2%  ~ (p=0.878 n=8)

       │  old.txt   │          new-struct.txt           │
       │    B/op    │    B/op     vs base               │
ECDH-8   40.00 ± 0%   32.00 ± 0%  -20.00% (p=0.000 n=8)

       │  old.txt   │          new-struct.txt           │
       │ allocs/op  │ allocs/op   vs base               │
ECDH-8   2.000 ± 0%   1.000 ± 0%  -50.00% (p=0.000 n=8)
```

### DIFF
--- a/goopenssl.h
+++ b/goopenssl.h
@@ -82,8 +82,9 @@ go_hash_sum(GO_EVP_MD_CTX_PTR ctx, GO_EVP_MD_CTX_PTR ctx2, unsigned char *out)
     return go_openssl_EVP_DigestFinal(ctx2, out, NULL);
 }
 
-// These wrappers allocate out_len on the C stack to avoid having to pass a pointer from Go, which would escape to the heap.
+// These wrappers allocate length variables on the C stack to avoid having to pass a pointer from Go, which would escape to the heap.
 // Use them only in situations where the output length can be safely discarded.
+
 static inline int
 go_openssl_EVP_EncryptUpdate_wrapper(GO_EVP_CIPHER_CTX_PTR ctx, unsigned char *out, const unsigned char *in, int in_len)
 {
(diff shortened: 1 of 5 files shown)


goopenssl.h
```diff
@@ -105,6 +106,48 @@ go_openssl_EVP_CipherUpdate_wrapper(GO_EVP_CIPHER_CTX_PTR ctx, unsigned char *ou
     return go_openssl_EVP_CipherUpdate(ctx, out, &len, in, in_len);
 }
 
+static inline int
+go_openssl_EVP_PKEY_get_raw_public_key_wrapper(const GO_EVP_PKEY_PTR pkey, unsigned char *pub, size_t len)
```
derekparker, line 110: These wrappers introduce a potential bug. When `len` is copied to the C stack we take the address of it, but we only have a reference to the value on the C stack, so the calling Go code does not receive the updated value, should it change.

derekparker (approved):
lgtm

gdams (approved):

qmuntal (approved):

ueno (approved):

PR #132, opened 2023-12-21
Add NewGCMTLS13 for Go TLS 1.3 support

dagood:
Implements `NewGCMTLS13` by extending `NewGCMTLS` and incorporating the logic described in the boringssl implementation at https://github.com/google/boringssl/blob/d1831d78c867ba51b3992ccc213fd201d2f4b0f1/crypto/fipsmodule/cipher/e_aes.c#L1397-L1403.

https://github.com/golang/go/commit/4106de901a8efe914cda6f6c4e8d45ff8c115da4 is in `master` and the `go1.22rc1` tag, and it depends on a new `NewGCMTLS13` function in the boring backend. I found out about this when it broke our upstream sync CI. It adds crypto/internal/boring `NewGCMTLS13`, calling `_goboringcrypto_EVP_aead_aes_128_gcm_tls13`. It didn't look all that complicated once I got to the boringssl implementation, so I thought I would try writing this PR.

---

The code comment I added is a little long-winded. It took me a bit to understand how the RFC applied. Maybe it's easier to see for people who are already more familiar with how these funcs are used, and the comment can be cut down.

### DIFF
--- a/aes.go
+++ b/aes.go
@@ -47,6 +47,12 @@ func NewGCMTLS(c cipher.Block) (cipher.AEAD, error) {
 	return c.(*aesCipher).NewGCMTLS()
 }
 
+// NewGCMTLS13 returns a GCM cipher specific to TLS 1.3 and should not be used
+// for non-TLS purposes.
+func NewGCMTLS13(c cipher.Block) (cipher.AEAD, error) {
+	return c.(*aesCipher).NewGCMTLS13()
+}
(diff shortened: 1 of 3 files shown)


ueno (approved):
Looks good to me!

qmuntal (approved):
LGTM!

gdams (approved):

PR #135, opened 2024-01-12
Add support for crypto/dsa

qmuntal:
This PR implements the necessary functions to support [crypto/dsa](https://pkg.go.dev/crypto/dsa).


xnox:
Do we care, given DSA is now legacy and unapproved for most usecases?

shims.h
```diff
@@ -19,6 +19,7 @@ enum {
     GO_EVP_PKEY_TLS1_PRF = 1021,
     GO_EVP_PKEY_HKDF = 1036,
     GO_EVP_PKEY_ED25519 = 1087,
+    GO_EVP_PKEY_DSA = 116,
     /* This is defined differently in OpenSSL 3 (1 << 11), but in our
      * code it is only used in OpenSSL 1.
     */
```
karianna, line 25:
```suggestion
     */
```

port_dsa.c
```diff
@@ -0,0 +1,83 @@
...
+    long _ignored1;
+    int _ignored2;
+    GO_BIGNUM_PTR p;
+    GO_BIGNUM_PTR q;
+    GO_BIGNUM_PTR g;
+    GO_BIGNUM_PTR pub_key;
+    GO_BIGNUM_PTR priv_key;
+};
+
+
```
karianna, line 19:
```suggestion
```

port_dsa.c
```diff
@@ -0,0 +1,83 @@
+// The following is a partial backport of crypto/dsa/dsa_pmeth.c,
+// commit cbc8a839959418d8a2c2e3ec6bdf394852c9501e on the
+// OpenSSL_1_1_0-stable branch.  The ctrl function has been removed.
```
karianna, line 3: Why was the ctrl function removed?  Not needed for our use case?

I assume the license of the openssl code is compatible for us to include this backport in this repo?

dsa_test.go
```diff
@@ -0,0 +1,205 @@
...
+	priv, err := openssl.NewPrivateKeyDSA(params, X, Y)
+	if err != nil {
+		t.Fatalf("error generating key: %s", err)
+	}
+
+	testDSASignAndVerify(t, priv)
+}
+
+func TestDSANewPrivateKeyWithDegenerateKeys(t *testing.T) {
+	// Signing with degenerate private keys should not cause an infinite
```
karianna, line 150:
```suggestion
	// Signing with degenerate private keys should not cause an infinite loop
```

dsa_test.go
```diff
@@ -0,0 +1,205 @@
...
+	// loop.
```
karianna, line 151:
```suggestion
```

dsa_test.go
```diff
@@ -0,0 +1,205 @@
...
+	priv, err := openssl.GenerateKeyDSA(params)
+	if err != nil {
+		t.Errorf("error generating key: %s", err)
+		return
+	}
+
+	testDSASignAndVerify(t, priv)
+}
+
+func testDSASignAndVerify(t *testing.T, priv *openssl.PrivateKeyDSA) {
```
karianna, line 77: Should we test Sign separately from verify or would that never make sense?

dsa_test.go
```diff
@@ -0,0 +1,205 @@
...
+		t.Error(err)
+		return
+	}
+	if !openssl.VerifyDSA(pub, hashed[:], sig) {
+		t.Error("compat: OpenSSL can't verify crypto/dsa signature")
+		return
+	}
+}
+
+func TestDSASignAndVerify(t *testing.T) {
```
karianna, line 130: My lack of Go is going to be exposed here :-).  Why the repeated function name with starting with a capital T?

dsa.go
```diff
@@ -0,0 +1,333 @@
...
+		C.go_openssl_DSA_free(dsa)
+		C.go_openssl_EVP_PKEY_free(pkey)
+		return nil, newOpenSSLError("EVP_PKEY_assign failed")
+	}
+	return pkey, nil
+}
+
+func newDSA3(params DSAParameters, X, Y BigInt) (C.GO_EVP_PKEY_PTR, error) {
+	if vMajor != 3 {
+		panic("incorrect vMajor version")
```
karianna, line 251: worth printing out what the vMajor variable is?

karianna (commented):
Some Q's because I don't understand Go that well :-) and some suggested nitpicks.

port_dsa.c
qmuntal: > Why was the ctrl function removed? Not needed for our use case?

This comment is outdated. I've updated it with something more meaningful.

> I assume the license of the openssl code is compatible for us to include this backport in this repo?

Not sure about this. This backporting approach is taken from the [.NET Runtime](https://github.com/dotnet/runtime/blob/main/src/native/libs/System.Security.Cryptography.Native/openssl_1_0_structs.h), and its third party notices doesn't mention OpenSSL, so I guess we are fine. My take is that these are just some getters and setters adapted for our need, nothing really interesting that worth mentioning, but I could be wrong. Let's talk about this in our next meeting.

dsa_test.go
qmuntal: For the common cases it makes sense to exercise both in the same test, else we would have to hardcode the test vectors inputs, which is always more obscure. Note that we do hardcode inputs and tests just the signing or the verification function for some degenerated cases, see `TestDSANewPrivateKeyWithDegenerateKeys` and `TestDSANewPublicKeyWithBadPublicKey`.

At the end is a matter of taste. Here I'm following the idiosyncrasy of the Go standard library, which prefers tests covering multiple function calls (almost like integration tests) over small and atomic unit tests. IMO it makes sense to follow this approach here too, as most of the tests are inspired by tests in the standard library (for example, these tests are very similar to the ones in https://github.com/golang/go/blob/a9e6a96ac092cc5191d759488ee761e9a403ab8f/src/crypto/dsa/dsa_test.go).

dsa_test.go
qmuntal: It's not a rookie question if you knew that the Go test framework requires test functions to start with the `Test` prefix (with capital T) and have the `(t *testing.T)` signature, else they won't be identified as such and won't be executed when running `go test` (see https://pkg.go.dev/testing). I'll assume you knew that 😄.

So, to your question, `testDSASignAndVerify` (with lowercase T) is just a reusable function that does what says on the tin: test that a given DSA key can be used to sign and verify. Note that this function is called from multiple places, as there are multiple tests that exercise larger workflows. On the other hand, `TestDSASignAndVerify` just exercise the DSA sign/verify workflow using hardcoded inputs, so it is a "small" wrapper around `testDSASignAndVerify`.

karianna (approved):

dsa.go
```diff
@@ -0,0 +1,331 @@
...
+	return DSAParameters{
+		P: bnToBig(p),
+		Q: bnToBig(q),
+		G: bnToBig(g),
+	}, nil
+}
+
+// NewPrivateKeyDSA creates a new DSA private key from the given parameters.
+func NewPrivateKeyDSA(params DSAParameters, X, Y BigInt) (*PrivateKeyDSA, error) {
+	if X == nil || Y == nil {
```
derekparker, line 119: nit: why CAPS for the `X` `Y` param names?

dsa.go
```diff
@@ -0,0 +1,331 @@
...
+		return nil, err
+	}
+	k := &PrivateKeyDSA{params, X, Y, pkey}
+	runtime.SetFinalizer(k, (*PrivateKeyDSA).finalize)
+	return k, nil
+}
+
+// NewPublicKeyDSA creates a new DSA public key from the given parameters.
+func NewPublicKeyDSA(params DSAParameters, Y BigInt) (*PublicKeyDSA, error) {
+	if Y == nil {
```
derekparker, line 133: nit: same as above regarding uppercase naming.

dsa.go
```diff
@@ -0,0 +1,331 @@
...
+	switch vMajor {
+	case 1:
+		dsa := getDSA(pkey)
+		if vMinor == 0 {
+			C.go_openssl_DSA_get0_key_backport(dsa, &y, &x)
+		} else {
+			C.go_openssl_DSA_get0_key(dsa, &y, &x)
+		}
+	case 3:
+		defer func() {
```
derekparker, line 161: Same question regarding freeing the params here and not the other block. Is this because the `DSAParameters` variable gets free'd itself anyways?

dsa.go
```diff
@@ -0,0 +1,331 @@
...
+func SignDSA(priv *PrivateKeyDSA, hash []byte) ([]byte, error) {
+	return evpSign(priv.withKey, 0, 0, 0, hash)
+}
+
+// VerifyDSA verifiessig using the public key, pub.
+func VerifyDSA(pub *PublicKeyDSA, hash []byte, sig []byte) bool {
+	return evpVerify(pub.withKey, 0, 0, 0, sig, hash) == nil
+}
+
+func newDSA(params DSAParameters, X, Y BigInt) (C.GO_EVP_PKEY_PTR, error) {
```
derekparker, line 187: (blanket nitpick) Why the uppercase input parameter naming?

dsa.go
```diff
@@ -0,0 +1,331 @@
...
+	case 1:
+		return newDSA1(params, X, Y)
+	case 3:
+		return newDSA3(params, X, Y)
+	default:
+		panic(errUnsupportedVersion())
+	}
+}
+
+func newDSA1(params DSAParameters, X, Y BigInt) (C.GO_EVP_PKEY_PTR, error) {
```
derekparker, line 198: Would it be better to implement this in C directly and call it once via CGO as opposed to doing a bunch of CGO calls and paying the cost of the context switching?

dsa.go
```diff
@@ -0,0 +1,331 @@
...
+	}
+	if C.go_openssl_EVP_PKEY_assign(pkey, C.GO_EVP_PKEY_DSA, unsafe.Pointer(dsa)) != 1 {
+		C.go_openssl_DSA_free(dsa)
+		C.go_openssl_EVP_PKEY_free(pkey)
+		return nil, newOpenSSLError("EVP_PKEY_assign failed")
+	}
+	return pkey, nil
+}
+
+func newDSA3(params DSAParameters, X, Y BigInt) (C.GO_EVP_PKEY_PTR, error) {
```
derekparker, line 248: Same question regarding implementing this in C and calling it once from Go.

derekparker (changes requested):

derekparker:
@ueno would be great to have your eyes on this as well.

dsa.go
qmuntal: Lapsus, will fix.

dsa.go
qmuntal: Lapsus, will fix.

dsa.go
```diff
@@ -0,0 +1,331 @@
...
+	case 1:
+		dsa := getDSA(pkey)
+		if vMinor == 0 {
+			C.go_openssl_DSA_get0_pqg_backport(dsa, &p, &q, &g)
+		} else {
+			C.go_openssl_DSA_get0_pqg(dsa, &p, &q, &g)
+		}
+	case 3:
+		defer func() {
+			C.go_openssl_BN_free(p)
```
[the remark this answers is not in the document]
qmuntal: Nop, we use a `DSA_get0_*` function to get them, and by convention all OpenSSL `get0` getters don't allocate new memory. From [DSA_get0_pqg](https://docs.openssl.org/1.1.1/man3/DSA_get0_pqg/#name) docs:

> These point directly to the internal representations of the values and therefore should not be freed directly.

dsa.go
qmuntal: I've done some profiling and that could make `NewPrivateKeyDSA` and `NewPublicKeyDSA` around 30% faster, from 3µ to 2µ. Note that 3 microseconds is already pretty damn fast, so doesn't worth porting it to C.

Disclaimer: I'm not opposed to batching C calls into single CGO function where profiling show relevant improvements. See for example [go_openssl_EVP_CIPHER_CTX_open_wrapper](https://github.com/golang-fips/openssl/blob/2ae7b9813caf936f33771cec1a4d07552295cbfa/goopenssl.h#L194).

dsa.go
```diff
@@ -0,0 +1,331 @@
...
+	"runtime"
+	"unsafe"
+)
+
+var (
+	paramPbits = C.CString("pbits")
+	paramQbits = C.CString("qbits")
+	paramP     = C.CString("p")
+	paramQ     = C.CString("q")
+	paramG     = C.CString("g")
```
dagood, line 17: Should these names differentiate that they're for DSA? E.g., we also have `paramRSA_Q` which is `rsa-factor2`. If a future PR accidentally uses `paramQ` in an RSA context, it would be easier to spot (or wouldn't occur) if this were called `paramDSA_Q`.

On the other hand, since it's literally `"q"`, maybe it should be intentionally shared if need be, and be moved to a different file than `dsa.go` if that happens?

dsa.go
```diff
@@ -0,0 +1,331 @@
...
+	defer runtime.KeepAlive(k)
+	return f(k._pkey)
+}
+
+// PublicKeyDSA represents a DSA public key.
+type PublicKeyDSA struct {
+	DSAParameters
+	Y BigInt
+
+	_pkey C.GO_EVP_PKEY_PTR
```
dagood, line 47: Include note found for this field in `rsa.go` and `ecdsa.go` (although admittedly not everywhere)?

```suggestion
	// _pkey MUST NOT be accessed directly. Instead, use the withKey method.
	_pkey C.GO_EVP_PKEY_PTR
```

dsa.go
```diff
@@ -0,0 +1,331 @@
...
+type DSAParameters struct {
+	P, Q, G BigInt
+}
+
+// PrivateKeyDSA represents a DSA private key.
+type PrivateKeyDSA struct {
+	DSAParameters
+	X, Y BigInt
+
+	_pkey C.GO_EVP_PKEY_PTR
```
dagood, line 30:
```suggestion
	// _pkey MUST NOT be accessed directly. Instead, use the withKey method.
	_pkey C.GO_EVP_PKEY_PTR
```

openssl.go
```diff
@@ -77,6 +77,13 @@ func errUnsupportedVersion() error {
 	return errors.New("openssl: OpenSSL version: " + utoa(vMajor) + "." + utoa(vMinor) + "." + utoa(vPatch))
 }
 
+// checkMajorVersion panics if the current major version is not expected.
+func checkMajorVersion(expected uint) {
+	if vMajor != expected {
+		panic("openssl: incorrect major version (" + strconv.Itoa(int(expected)) + "), expected " + strconv.Itoa(int(expected)))
```
dagood, line 83: Wrong var:

```suggestion
		panic("openssl: incorrect major version (" + strconv.Itoa(int(vMajor)) + "), expected " + strconv.Itoa(int(expected)))
```

dagood (commented):
A few surface-level suggestions.

dsa.go
qmuntal: This raises a valid point. I'll switch the naming pattern to use the names given in the OpenSSL source code (e.g. `OSSL_PKEY_PARAM_FFC_PBITS` instead of `pbits`. Node that DSA parameters are shared with other FFC algorithms, but we don't implement more of those yet, so we can keep this variable here for the moment.

I'll update the parameter names for other algorithms (ECDSA, HMAC and RSA) in a separate PR.

dsa.go
```diff
@@ -0,0 +1,333 @@
...
+			return nil, newOpenSSLError("DSA_generate_key failed")
+		}
+	}
+	pkey := C.go_openssl_EVP_PKEY_new()
+	if pkey == nil {
+		C.go_openssl_DSA_free(dsa)
+		return nil, newOpenSSLError("EVP_PKEY_new failed")
+	}
+	if C.go_openssl_EVP_PKEY_assign(pkey, C.GO_EVP_PKEY_DSA, unsafe.Pointer(dsa)) != 1 {
+		C.go_openssl_DSA_free(dsa)
```
dagood, line 243: This long lifetime is causing a decent amount of duplication... might be worth replacing the current frees with this pattern (also used by `(k *PrivateKeyECDH) PublicKey()`):

```go
	dsa := C.go_openssl_DSA_new()
	defer func() {
		C.go_openssl_DSA_free(dsa)
	}()
//...
	dsa = nil // Ownership has been transferred to pkey.
	return pkey, nil
```

dsa.go
```diff
@@ -0,0 +1,333 @@
...
+	p, q, g := bigToBN(params.P), bigToBN(params.Q), bigToBN(params.G)
+	defer func() {
+		C.go_openssl_BN_free(p)
+		C.go_openssl_BN_free(q)
+		C.go_openssl_BN_free(g)
+	}()
+	if C.go_openssl_OSSL_PARAM_BLD_push_BN(bld, OSSL_PKEY_PARAM_FFC_P, p) != 1 ||
+		C.go_openssl_OSSL_PARAM_BLD_push_BN(bld, OSSL_PKEY_PARAM_FFC_Q, q) != 1 ||
+		C.go_openssl_OSSL_PARAM_BLD_push_BN(bld, OSSL_PKEY_PARAM_FFC_G, g) != 1 {
+		return nil, newOpenSSLError("OSSL_PARAM_BLD_push_BN")
```
dagood, line 267: Nit: newline to clarify where the wrapped `if` stops? (Not currently a bad case though, because so much repeats.)

```suggestion
		C.go_openssl_OSSL_PARAM_BLD_push_BN(bld, OSSL_PKEY_PARAM_FFC_G, g) != 1 {

		return nil, newOpenSSLError("OSSL_PARAM_BLD_push_BN")
```

dsa.go
```diff
@@ -0,0 +1,333 @@
...
+	pkey, err := newEvpFromParams(C.GO_EVP_PKEY_DSA, selection, bldparams)
+	if err != nil {
+		return nil, err
+	}
+	if y != nil {
+		return pkey, nil
+	}
+	// pkey doesn't contain the public component, but the crypto/dsa expects
+	// it to be always there. Generate a new key using pkey as domain parameters
+	// placeholder.
```
dagood, line 301:
```suggestion
	// pkey doesn't contain the public component, but the crypto/dsa package
	// expects it to be always there. Generate a new key using pkey as domain
	// parameters placeholder.
```

dagood (approved):

dsa.go
qmuntal: Like the idea, thanks.

dsa.go
```diff
@@ -225,22 +233,18 @@ func newDSA1(params DSAParameters, x, y BigInt) (C.GO_EVP_PKEY_PTR, error) {
...
 		}
 	}
-	pkey := C.go_openssl_EVP_PKEY_new()
+	pkey = C.go_openssl_EVP_PKEY_new()
 	if pkey == nil {
-		C.go_openssl_DSA_free(dsa)
 		return nil, newOpenSSLError("EVP_PKEY_new failed")
 	}
 	if C.go_openssl_EVP_PKEY_assign(pkey, C.GO_EVP_PKEY_DSA, unsafe.Pointer(dsa)) != 1 {
-		C.go_openssl_DSA_free(dsa)
```
dagood: Assuming the existing code was right and calling `C.go_openssl_DSA_free(dsa)` is needed in this error case, this seems wrong.

My understanding is that `EVP_PKEY_assign` is what actually transfers ownership, and when we hit this `!= 1` the ownership isn't transferred, and that's why `C.go_openssl_DSA_free(dsa)` is needed. But at this point, `pkey` isn't `nil`, so the deferred cleanup won't call `DSA_free`.

(This is something I liked about the explicit `dsa = nil // Ownership has been transferred to pkey.`--it's very clear what's intended and doesn't tie state together.)

dagood (changes requested):

dsa.go
dagood: Oop, I missed how the change to use a named return affects this. Never mind!

dagood (approved):

dsa.go
```diff
@@ -0,0 +1,338 @@
...
+	runtime.SetFinalizer(k, (*PrivateKeyDSA).finalize)
+	return k, nil
+}
+
+// SignDSA signs a hash (which should be the result of hashing a larger message).
+func SignDSA(priv *PrivateKeyDSA, hash []byte) ([]byte, error) {
+	return evpSign(priv.withKey, 0, 0, 0, hash)
+}
+
+// VerifyDSA verifiessig using the public key, pub.
```
ueno, line 184: "verifies sig" ?

ueno (approved):
Looks good to me.

karianna:
@derekparker Can we get a re-review when you get a chance?

derekparker (approved):
LGTM

PR #136, opened 2024-02-13
rsa: Allow omission of precomputed values in NewPrivateKeyRSA

ueno:
OpenSSL 3.0 and 3.1 required precomputed values for CRT (Dp, Dq, and Qinv), when the primes (P and Q) are included in the params array.

Since all of them can be derived from N, E, and D, this patch simply stops passing those values if any of them are missing.

See also: https://github.com/openssl/openssl/pull/22334

### DIFF
--- a/rsa.go
+++ b/rsa.go
@@ -384,14 +384,30 @@ func newRSAKey3(isPriv bool, N, E, D, P, Q, Dp, Dq, Qinv BigInt) (C.GO_EVP_PKEY_
 		return nil, newOpenSSLError("OSSL_PARAM_BLD_new")
 	}
 	defer C.go_openssl_OSSL_PARAM_BLD_free(bld)
-	var comps = [...]struct {
+
+	type bigIntParam struct{
 		name *C.char
 		num  BigInt
-	}{
-		{paramRSA_N, N}, {paramRSA_E, E}, {paramRSA_D, D},
-		{paramRSA_P, P}, {paramRSA_Q, Q},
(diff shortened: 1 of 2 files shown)


rsa.go
```diff
@@ -384,14 +384,28 @@ func newRSAKey3(isPriv bool, N, E, D, P, Q, Dp, Dq, Qinv BigInt) (C.GO_EVP_PKEY_
...
+
+	// OpenSSL 3.0 and 3.1 required all the precomputed values if
+	// P and Q are present. See:
+	// https://github.com/openssl/openssl/pull/22334
+	if P != nil && Q != nil && Dp != nil && Dq != nil && Qinv != nil {
+		precomputed := [...]bigIntParam{
+			{paramRSA_P, P}, {paramRSA_Q, Q},
+			{paramRSA_Dp, Dp}, {paramRSA_Dq, Dq}, {paramRSA_Qinv, Qinv},
+		}
+		comps = append(comps, precomputed[:]...)
```
qmuntal, line 406: This will have to grow the slice, as `comps` length is `3` at this point, producing an allocation. Do something like this when initializing `comps` to avoid the allocation:

```go
comps := make([]bigIntParam, 0, 8)
comps = append(comps, required...)
```

qmuntal (changes requested):

rsa.go
ueno: Thanks; rewritten in that way.

qmuntal (approved):

derekparker (approved):
lgtm

PR #138, opened 2024-02-20
Fix loop capture bug in parallel test

derekparker:


openssl/rsa_test.go
```diff
@@ -175,7 +175,8 @@ func TestRSAKeyGeneration(t *testing.T) {
 	for _, size := range []int{2048, 3072} {
 		t.Run(strconv.Itoa(size), func(t *testing.T) {
 			t.Parallel()
-			priv, pub := newRSAKey(t, size)
+			sz := size
+			priv, pub := newRSAKey(t, sz)
```
dagood, line 179: I don't think this solves the problem: at this point, making the `sz := size` copy is too late--the `for` loop has already continued to future iterations because of `t.Parallel`. Putting the copy before `t.Run` is what makes sense to me.

`size := size` also seems more "normal" to me in Go, because nobody should be accessing the original `size`.

derekparker: Ack good catch, fixed.

dagood (approved):

gdams (approved):

PR #143, opened 2024-04-11
[go1.20-support] Stop using deprecated EC functions for ECDSA

ueno:
This backports #58 to the `go1.20-support` branch, to address golang-fips/go#177.


derekparker (approved):
LGTM

PR #144, opened 2024-05-09
[go1.20-support] Stop using deprecated RSA functions

ueno:
This backports https://github.com/golang-fips/openssl/pull/57 to the go1.20-support branch, similar to #143.


karianna:
@ueno Is this PR still valid? If so I'll ask some of the MSFT folks to review.

PR #145, opened 2024-05-10
Protect GO_EVP_CIPHER_CTX_PTR with sync.Mutex

lyoung-confluent:
See https://github.com/golang-fips/go/issues/187 for details on why this is necessary. Protecting these with a mutex is a brute-force approach and presumably has negative performance implications, another option could be to go back to creating a `GO_EVP_CIPHER_CTX_PTR` within each encrypt/decrypt operation instead of caching the context.

### DIFF
--- a/cipher.go
+++ b/cipher.go
@@ -145,7 +145,9 @@ func loadCipher(k cipherKind, mode cipherMode) (cipher C.GO_EVP_CIPHER_PTR) {
 
 type evpCipher struct {
 	key       []byte
+	encLock   sync.Mutex
 	enc_ctx   C.GO_EVP_CIPHER_CTX_PTR
+	decLock   sync.Mutex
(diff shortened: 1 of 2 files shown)


derekparker:
Thanks for sending this fix!

There's definitely going to be a performance hit when trying to run these operations in parallel. My gut instinct is instead of using this mutex we instead create a new `EVP_CIPHER_CTX` for each operation instead of reusing, via `newCipherCtx` and don't store it in any of the structs such as `evpCipher`. This has the drawback that the caller will have to free that context to avoid leaks, but that's ok. We can also wrap the `EVP_CIPHER_CTX` in a struct with a finalizer to free the memory when that object is GC'd in the case of a programmer error where free is not called and it starts to leak.

cc @dagood @qmuntal @ueno @simo5

derekparker:
Closing in favor of https://github.com/golang-fips/openssl/pull/146.

PR #146, opened 2024-05-20
cipher: do not reuse cipher ctx for certain operations

derekparker:
Fixes https://github.com/golang-fips/go/issues/187

CBC and CTR are left as-is, as we also cached the context in the v1 version, so technically not a regression.

### DIFF
--- a/aes_test.go
+++ b/aes_test.go
@@ -421,14 +421,6 @@ func TestDecryptInvariantReusableNonce(t *testing.T) {
 	testDecrypt(t, true)
 }
 
-func Test_aesCipher_finalize(t *testing.T) {
-	// Test that aesCipher.finalize does not panic if neither Encrypt nor Decrypt have been called.
-	// This test is important because aesCipher.finalize contains logic that is normally not exercided while testing.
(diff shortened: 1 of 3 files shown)


derekparker:
Tested locally against Go 1.22.3 (golang-fips/go tip).

ueno (approved):
LGTM

derekparker:
@lyoung-confluent I've tested against your reproducer but could you verify this fix works for you as well?

Also @dagood or @qmuntal if you can take a look as well that would be greatly appreciated.

qmuntal:
This looks like an important issue to fix. @dagood we should port this PR to the CNG backend, ideally before go1.23 is released.

Can't properly review it myself, I'm on paternity leave. Will defer the review to @dagood. Fyi @karianna.

dagood:
Thanks @qmuntal for the ping, and especially the note about CNG. This is definitely near the top of my TODO list, sorry it's been hanging a bit.

@karianna offered to help with an initial review for the PR on behalf the Microsoft team to get this through, so I've added him as a collaborator.

karianna (approved):
LGTM

PR #150, opened 2024-08-29
V1 fixes

dbenoit17:
* Add math/big import in notboring.go
* Use legacy EVP_MD_CTX_md in openssl 1.1

### DIFF
--- a/openssl/openssl_port_hmac.c
+++ b/openssl/openssl_port_hmac.c
@@ -14,7 +14,12 @@ DEFINEFUNCINTERNAL(EVP_PKEY *,
 		   (int type, ENGINE *e, const unsigned char *key, int keylen),
 		   (type, e, key, keylen))
 DEFINEFUNCINTERNAL(int, EVP_MD_CTX_reset, (EVP_MD_CTX *ctx), (ctx))
+
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
 DEFINEFUNCINTERNAL(const EVP_MD *, EVP_MD_CTX_get0_md, (const EVP_MD_CTX *ctx), (ctx))
+#else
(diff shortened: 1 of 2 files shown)


openssl/openssl_port_hmac.c
```diff
@@ -14,7 +14,12 @@ DEFINEFUNCINTERNAL(EVP_PKEY *,
 		   (int type, ENGINE *e, const unsigned char *key, int keylen),
 		   (type, e, key, keylen))
 DEFINEFUNCINTERNAL(int, EVP_MD_CTX_reset, (EVP_MD_CTX *ctx), (ctx))
+
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
```
karianna, line 18: Is that a magic number we can extract as a global named const?

simo5: you mean somethign like:
```
#define OPENSSL_3_0_VERSION 0x30000000L
#if OPENSSL_VERSION_NUMBER >= OPENSSL_3_0_VERSION
```
?

ueno (approved):

openssl/openssl_port_hmac.c
ueno: Maybe we could add a compatibility definition of `OPENSSL_VERSION_PREREQ` (added in 1.1.1) so we can check the version with `#if OPENSSL_VERSION_REREQ(3,0) ...` ?

dbenoit17: I think using a macro or constant could be more intuitive given that we only discern between a handful of openssl versions.  But let's merge this and start a follow up conversation given the existing precedent for using numeric literals.

karianna (approved):

openssl/openssl_port_hmac.c
karianna: Extracted to a new issue #151

PR #161, opened 2024-09-03
Only implement encoding.BinaryMarshaler for hashers provided by built-in providers

### LINKED ISSUE #156: MarshalBinary and UnmarshalBinary can't be used with custom providers
We currently implement [encoding.BinaryMarshaler](https://pkg.go.dev/encoding#BinaryMarshaler) and [encoding.BinaryUnmarshaler](https://pkg.go.dev/encoding#BinaryUnmarshaler) for the hashers returned by `openssl.NewMD5`, `openssl.NewSHA1`, `openssl.NewSHA224`, `openssl.NewSHA256`, `openssl.NewSHA384`, and `openssl.NewSHA512`.

The marshaling implementation for OpenSSL 3 assumes that all OpenSSL providers will have the same memory layout as the build-in providers: 

https://github.com/golang-fips/openssl/blob/f8eea431f613cac3064280dbb97d9ce1a39de973/hash.go#L202-L224)

The problem is that most providers will have their own memory layout for the aforementioned hashers. This means that our marshaling implementation only works when using the built-in providers.

I don't have a good solution for this issue yet.

qmuntal:
We can only marshal/unmarshal the internal state of hashers provided by built-in providers, which are the only ones from which we know the memory layout.

This PR does some struct embedding tricks so the hashers returned by`NewMD5`, `NewSHA1`, `NewSHA224`, `NewSHA256`, `NewSHA384`, and `NewSHA512`, only implement the `encoding.BinaryMarshaler` and `encoding.BinaryUnmarshaler` interface if the hashers are provided by the fips or the default provider.

Fixes #156.

### DIFF
--- a/hash.go
+++ b/hash.go
@@ -10,6 +10,7 @@ import (
 	"hash"
 	"runtime"
 	"strconv"
+	"sync"
 	"unsafe"
 )
 
@@ -108,6 +109,37 @@ func SHA3_512(p []byte) (sum [64]byte) {
 		panic("openssl: SHA3_512 failed")
 	}
 	return
+}
+
+var isMarshallableMap sync.Map
+
+// isHashMarshallable returns true if its memory layout
+// is known by this library, therefore it can be marshalled.
+func isHashMarshallable(cb crypto.Hash) bool {
+	if vMajor == 1 {
+		return true
+	}
+	if v, ok := isMarshallableMap.Load(cb); ok {
+		return v.(bool)
+	}
+	md := cryptoHashToMD(cb)
+	if md == nil {
+		return false
+	}
+	prov := C.go_openssl_EVP_MD_get0_provider(md)
+	if prov == nil {
+		return false
+	}
+	cname := C.go_openssl_OSSL_PROVIDER_get0_name(prov)
+	if cname == nil {
+		return false
+	}
+	name := C.GoString(cname)
+	// We only known the memory layout of the built-in providers.
+	// See evpHash.hashState for more details.
+	marshallable := name == "default" || name == "fips"
+	isMarshallableMap.Store(cb, marshallable)
+	return marshallable
 }
 
 // evpHash implements generic hash methods.
@@ -244,9 +276,11 @@ func (h *md4Hash) Sum(in []byte) []byte {
 
 // NewMD5 returns a new MD5 hash.
 func NewMD5() hash.Hash {
-	return &md5Hash{
-		evpHash: newEvpHash(crypto.MD5, 16, 64),
+	h := newEvpHash(crypto.MD5, 16, 64)
+	if isHashMarshallable(crypto.MD5) {
+		return &md5Marshal{md5Hash{evpHash: h}}
 	}
+	return &md5Hash{evpHash: h}
 }
 
 // md5State layout is taken from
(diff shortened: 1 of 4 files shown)


derekparker (approved):
LGTM

hash.go
```diff
@@ -110,6 +111,37 @@ func SHA3_512(p []byte) (sum [64]byte) {
 	return
 }
 
+var isMarshallableMap sync.Map
+
+// isHashMarshallable returns true if its memory layout
+// is known by this library, therefore it can be marshalled.
+func isHashMarshallable(cb crypto.Hash) bool {
```
dagood, line 118: Was `cb` meant to be `ch` (crypto hash)?

hash.go
```diff
@@ -407,9 +453,11 @@ func (h *sha224Hash) Sum(in []byte) []byte {
 
 // NewSHA256 returns a new SHA256 hash.
 func NewSHA256() hash.Hash {
-	return &sha256Hash{
-		evpHash: newEvpHash(crypto.SHA256, 256/8, 64),
+	h := newEvpHash(crypto.SHA256, 256/8, 64)
+	if isHashMarshallable(crypto.SHA256) {
+		return &sha256Marshal{sha256Hash{evpHash: h}}
 	}
+	return &sha256Hash{evpHash: h}
```
dagood, line 460: It looks like a tiny bit more in common could be pulled out:

```go
	h := sha256Hash{evpHash: newEvpHash(crypto.SHA256, 256/8, 64)}
	if isHashMarshallable(crypto.SHA256) {
		return &sha256Marshal{h}
	}
	return &h
```

(Unless there's a perf/rule-of-thumb reason not to?)

hash_test.go
```diff
@@ -41,30 +41,27 @@ func cryptoToHash(h crypto.Hash) func() hash.Hash {
...
+		crypto.SHA3_224,
+		crypto.SHA3_256,
+		crypto.SHA3_384,
+		crypto.SHA3_512,
 	}
-	for _, tt := range tests {
-		tt := tt
-		t.Run(tt.h.String(), func(t *testing.T) {
+	for _, cb := range tests {
+		cb := cb
```
dagood, line 58: Also `ch` maybe?

hash.go
qmuntal: Good shot. I like it 😸

hash.go
```diff
@@ -232,6 +235,9 @@ func (h *evpHash) sum(out []byte) {
 // The EVP_MD_CTX memory layout has changed in OpenSSL 3
 // and the property holding the internal structure is no longer md_data but algctx.
 func (h *evpHash) hashState() unsafe.Pointer {
+	if !h.marshallable {
+		panic("openssl: hash state is not marshallable")
```
dagood, line 239: Would this ever come up from user code or OpenSSL oddity, or would there have to be a typo+missed test in golang-fips/openssl?

Wondering if it might make more sense to return this as a second value from `newEvpHash` rather than store it in the struct (at the cost of not having this runtime check).

dagood (approved):

hash.go
qmuntal: There would have to be a typo+missed test, aka it is a programmer error, not something that should happen at runtime.

I initially returned that property as second value from `newEvpHash`, but that made all `NewSHAX` a bit less cleaner, and having the runtime check is a nice bonus. Also, with the current implementation, `NewSHAX` functions are inlinable, which can help escape analysis take better decisions.

PR #162, opened 2024-09-04
Use valid address in pthread_setspecific to silence warnings

qmuntal:
Building with a newer glibc (probably since 2.34) causes the following warning:

```cmd
# vendor/github.com/golang-fips/openssl/v2
thread_setup_unix.c: In function 'thread_id':
thread_setup_unix.c:32:12: warning: 'pthread_setspecific' expecting 1 byte in a region of size 0 [-Wstringop-overread]
   32 |     (void) pthread_setspecific(destructor_key, (void*)1);
      |            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from thread_setup_unix.c:5:
/usr/lib/gcc/x86_64-pc-linux-gnu/11.2.0/include-fixed/pthread.h:1317:12: note: in a call to function 'pthread_setspecific' declared with attribute 'access (none, 2)'
 1317 | extern int pthread_setspecific (pthread_key_t __key,
      |            ^~~~~~~~~~~~~~~~~~~
```

The is caused by the `access` attribute annotation of `pthread_setspecific`, which was recently added.

GCC [defines](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html) the none mode of access as:

The access mode none specifies that the pointer to which it applies is not used to access the referenced object at all. Unless the pointer is null the pointed-to object must exist and have at least the size as denoted by the size-index argument. The object need not be initialized. The mode is intended to be used as a means to help validate the expected object size, for example in functions that call __builtin_object_size. See Object Size Checking.

We are passing `(void*)1` as a NULL pointer, which doesn't meet the `none` access requirements. This PR fixes that by using a static stub char pointer.

Fixes #1305.


thread_setup_unix.c
```diff
@@ -29,7 +32,7 @@ static void thread_id(GO_CRYPTO_THREADID_PTR tid)
     // least once on any thread with associated error state. The thread-local
     // variable needs to be set to a non-NULL value so that the destructor will
     // be called when the thread exits. The actual value does not matter.
-    (void) pthread_setspecific(destructor_key, (void*)1);
+    (void) pthread_setspecific(destructor_key, &stub);
```
karianna, line 35: I assume no backwards compatibility issue with older versions of gcc/glibc?

karianna (approved):

thread_setup_unix.c
qmuntal: `&stub` is a good old pointer, so nobody should complain about it. At least our CI is happy, which uses a glibc version that doesn't implement `access` modes.

derekparker (approved):
LGTM

thread_setup_unix.c
```diff
@@ -9,6 +9,9 @@ static pthread_mutex_t *mutex_buf = NULL;
 
 static pthread_key_t destructor_key;
 
+/* Used in pthread_setspecific. See https://github.com/microsoft/go/issues/1305. */
+static char stub;
```
dagood, line 13: Does this need to exist at the global scope, or could `static char stub;` or `char stub;` be declared just before the call?

If there's no difference, I think putting the declaration inside `thread_id` would be a bit clearer. And IMO putting it on the stack (if possible) raises the fewest questions from a reader's perspective. 😄

dagood (approved):

thread_setup_unix.c
qmuntal: Looks putting the variable in the function scope also works, and it does look clearer. I'll keep the `static` keyword, though, I would rather not pass a pointer to a stack-allocated variable that can be used outside the function frame, just in case it is used somewhere 😅.

PR #163, opened 2024-09-04
Only omit RSA primes if precomputed values are missing in OpenSSL 3.0 and 3.1

qmuntal:
Since OpenSSL 3.2, RSA primes can coexist with empty precomputed values, see https://github.com/openssl/openssl/pull/22334.
Some providers, such as SymCrypt, compute the RSA private key using the prime numbers `P` and `Q` instead of loading it from the private exponent `D`. So they always need `P` and `Q`, even if the precomputed values are omitted. This means that these providers are not fully compatible with OpenSSL 3.0 and 3.1, but they do work with OpenSSL 3.2 and higher.

This PR relaxes the restriction added in #136 to only take effect with OpenSSL 3.0 and 3.1.

### DIFF
--- a/rsa.go
+++ b/rsa.go
@@ -396,14 +396,15 @@ func newRSAKey3(isPriv bool, n, e, d, p, q, dp, dq, qinv BigInt) (C.GO_EVP_PKEY_
 	}
 	comps = append(comps, required[:]...)
 
-	// OpenSSL 3.0 and 3.1 required all the precomputed values if
-	// P and Q are present. See:
-	// https://github.com/openssl/openssl/pull/22334
-	if vMinor >= 2 || (p != nil && q != nil && dp != nil && dq != nil && qinv != nil) {
-		if p != nil && q != nil {
+	if p != nil && q != nil {
+		allPrecomputedExists := dp != nil && dq != nil && qinv != nil
+		// OpenSSL 3.0 and 3.1 required all the precomputed values if
+		// P and Q are present. If they are not, we need to omit also P and Q.
+		// See https://github.com/openssl/openssl/pull/22334
+		if vMinor >= 2 || allPrecomputedExists {
 			comps = append(comps, bigIntParam{OSSL_PKEY_PARAM_RSA_FACTOR1, p}, bigIntParam{OSSL_PKEY_PARAM_RSA_FACTOR2, q})
 		}
-		if dp != nil && dq != nil && qinv != nil {
+		if allPrecomputedExists {
 			comps = append(comps,
 				bigIntParam{OSSL_PKEY_PARAM_RSA_EXPONENT1, dp},
 				bigIntParam{OSSL_PKEY_PARAM_RSA_EXPONENT2, dq},
--- a/rsa.go
+++ b/rsa.go
@@ -399,7 +399,7 @@ func newRSAKey3(isPriv bool, n, e, d, p, q, dp, dq, qinv BigInt) (C.GO_EVP_PKEY_
 	// OpenSSL 3.0 and 3.1 required all the precomputed values if
 	// P and Q are present. See:
 	// https://github.com/openssl/openssl/pull/22334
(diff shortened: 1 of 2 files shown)


karianna (commented):
Can / Should we add a test for this?

rsa.go
```diff
@@ -397,15 +397,21 @@ func newRSAKey3(isPriv bool, N, E, D, P, Q, Dp, Dq, Qinv BigInt) (C.GO_EVP_PKEY_
...
-	// https://github.com/openssl/openssl/pull/22334
-	if P != nil && Q != nil && Dp != nil && Dq != nil && Qinv != nil {
-		precomputed := [...]bigIntParam{
-			{paramRSA_P, P}, {paramRSA_Q, Q},
-			{paramRSA_Dp, Dp}, {paramRSA_Dq, Dq}, {paramRSA_Qinv, Qinv},
+	if P != nil && Q != nil {
+		allPrecomputedExists := Dp != nil && Dq != nil && Qinv != nil
+		// OpenSSL 3.0 and 3.1 required all the precomputed values if
+		// P and Q are present. If they are not, we need to omit also P and Q.
+		// See https://github.com/openssl/openssl/pull/22334
```
dagood, line 404: I had a hard time with the past tense "required" vs. present "are", and "they", and what OpenSSL versions are being discussed. This is how I interpret it:

```suggestion
		// The precomputed values should only be passed if P and Q are present
		// and every precomputed value is present. (If any precomputed value is
		// missing, don't pass any of them.)
		//
		// In OpenSSL 3.0 and 3.1, we must also omit P and Q if any precomputed
		// value is missing. See https://github.com/openssl/openssl/pull/22334
```

qmuntal:
> Can / Should we add a test for this?

Good point. I've added some tests and, while here, improved a bit all the RSA tests.

karianna (approved):

rsa_test.go
```diff
@@ -13,11 +13,73 @@ import (
...
+			return newRSAKeyFromParams(t, N, E, D, P, Q, nil, Dq, Qinv)
+		}},
+		{"noDq", false, func() (*openssl.PrivateKeyRSA, *openssl.PublicKeyRSA) {
+			return newRSAKeyFromParams(t, N, E, D, P, Q, Dp, nil, Qinv)
+		}},
+		{"noQinv", false, func() (*openssl.PrivateKeyRSA, *openssl.PublicKeyRSA) {
+			return newRSAKeyFromParams(t, N, E, D, P, Q, Dp, Dq, nil)
+		}},
+		{"none", false, func() (*openssl.PrivateKeyRSA, *openssl.PublicKeyRSA) {
+			return newRSAKeyFromParams(t, N, E, D, P, Q, nil, nil, nil)
```
dagood, line 74: Maybe it's a bit silly because we know the internal logic is basically just `Dp && Dq && Qinv`, but this could be done exhaustively. (Iterate from `0b001` to `0b111` and skip passing the `1` bits, or something like that.)

dagood (approved):

karianna:
@qmuntal will need a rebase now

rsa_test.go
qmuntal: Doing an exhaustive test seems good, will do that. Won't use your bit test suggestion, though, it is too clever, I'll just hardcode the possible combinations 😄.

qmuntal:
@ueno can you take a look at this? Thanks.

ueno (approved):
LGTM, thanks!

rsa_test.go
dagood: Another way without as much cleverness (and with some nesting) would be:

```go
	falseTrue := []bool{true, false}
	for _, withDp := range falseTrue {
		for _, withDq := range falseTrue {
			for _, withQinv := range falseTrue {
				//...
			}
		}
	}
```

But still not quite as simple as table-driven possibilities, to be sure!

PR #164, opened 2024-09-04
Skip tests not supported by SymCrypt

qmuntal:
The SymCrypt provider does not currently support RSA encryption with a distinct MGF1 hash function nor using RSA PKCS1.5 with unhashed messages.

This PR skips the affected tests when the SymCrypt provider is available, which means that it will probably be used and the tests will fail.

### DIFF
--- a/openssl.go
+++ b/openssl.go
@@ -112,7 +112,7 @@ func FIPS() bool {
 }
 
 // isProviderAvailable checks if the provider with the given name is available.
-// This function is used in export_test.go, as test files can't access C functions.
+// This function is used in export_test.go, but must be defined here as test files can't access C functions.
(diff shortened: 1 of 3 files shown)


karianna (approved):

openssl.go
```diff
@@ -111,6 +111,14 @@ func FIPS() bool {
 	}
 }
 
+// isProviderAvailable checks if the provider with the given name is available.
+// This function is used in export_test.go, as test files can't access C functions.
```
dagood, line 115:
```suggestion
// This function is used in export_test.go, but must be defined here as test files can't access C functions.
```

dagood (approved):

PR #167, opened 2024-09-05
Implement Clone for all hashers

qmuntal:
In #161 I refactored our hash implementations to only implement the `encoding.BinaryMarshaler` interface when using the built-in providers. This raises a new problem, as the standard library uses that interface to clone a hash when doing a TLS 1.3 handshake: https://github.com/golang/go/blob/2707d42966f8985a6663c93e943b9a44b9399fca/src/crypto/tls/handshake_server_tls13.go#L436.

The issue can be fixed by providing a `Clone` method that is not provider-dependent and patch the standard library to use that instead. In fact, we are already following this approach with the Windows CNG backend in the Microsoft Go fork: https://github.com/microsoft/go/blob/294feac5a41d50e2ef261d6acca9feb017f873d0/patches/0005-Add-CNG-crypto-backend.patch#L1061-L1069. I plan to also use this new `Clone` method for the OpenSSL backend.

While here, refactor the hash tests so they are a bit more atomic.

### DIFF
--- a/hash.go
+++ b/hash.go
@@ -230,6 +230,9 @@ func (h *evpHash) sum(out []byte) {
 	runtime.KeepAlive(h)
 }
 
+// clone returns a new evpHash object that is a deep clone of itself.
+// The duplicate object contains all state and data contained in the
+// original object at the point of duplication.
 func (h *evpHash) clone() (*evpHash, error) {
 	ctx := C.go_openssl_EVP_MD_CTX_new()
 	if ctx == nil {
@@ -305,6 +308,9 @@ func (h *md4Hash) Sum(in []byte) []byte {
 	return append(in, h.out[:]...)
 }
 
+// Clone returns a new [hash.Hash] object that is a deep clone of itself.
+// The duplicate object contains all state and data contained in the
+// original object at the point of duplication.
 func (h *md4Hash) Clone() (hash.Hash, error) {
 	c, err := h.clone()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


hash.go
```diff
@@ -230,6 +230,31 @@ func (h *evpHash) sum(out []byte) {
 	runtime.KeepAlive(h)
 }
 
+func (h *evpHash) clone() (*evpHash, error) {
```
karianna, line 233: Maybe a comment here on the depth of the clone?  Is it shallow or deep? Likewise for other clones below

qmuntal: Make sense. I'll document that it is a deep clone.

hash_test.go
```diff
@@ -77,38 +78,115 @@ func TestHash(t *testing.T) {
...
+			}
+			h := cryptoToHash(ch)()
+			if _, ok := h.(encoding.BinaryMarshaler); !ok {
+				t.Skip("skipping: not supported")
+			}
+			_, err := h.Write(msg)
+			if err != nil {
+				t.Fatal(err)
+			}
+			h2, err := h.(interface{ Clone() (hash.Hash, error) }).Clone()
```
dagood, line 139: Maybe mention in a comment that this test is written in the same way that we expect the Clone method to be used in a Go fork (if it's necessary to use it)? Otherwise, it seems odd that we wouldn't use a named (and exported) interface here.

dagood (approved):

karianna (approved):

gdams (approved):

PR #169, opened 2024-09-06
Panic if DES encryption or decryption fails

qmuntal:
We missed to panic when a `desCipherWithoutCBC` encryption or decryption failed, which made our DES implementation not conformant with the [cipher.Block](https://pkg.go.dev/crypto/cipher#Block) interface, specially for overlapping inputs.

The fix is really small, see this commit: 9e68e597b7cac57a72912bfdae0e7ad49cada0b2.

The bulk of the changes in this repo are just adding tests to make sure that we don't regress.

Fixes https://github.com/microsoft/go/issues/1290.

While here, fix the 3DES CBC support detection logic, which was backwards.

This PR will have to be backported into the Microsoft Go 1.23 support branch.

### DIFF
--- a/block_test.go
+++ b/block_test.go
@@ -1,0 +1,261 @@
+package openssl_test
+
+import (
+	"bytes"
+	"crypto/cipher"
+	"io"
+	"math/rand"
+	"testing"
+	"time"
+)
+
+// This file is a copy of https://github.com/golang/go/blob/9e9b1f57c26a6d13fdaebef67136718b8042cdba/src/crypto/internal/cryptotest/block.go.
+
+type MakeBlock func(key []byte) (cipher.Block, error)
+
+// testBlock performs a set of tests on cipher.Block implementations, checking
+// the documented requirements of BlockSize, Encrypt, and Decrypt.
+func testBlock(t *testing.T, keySize int, mb MakeBlock) {
+	// Generate random key
+	key := make([]byte, keySize)
+	newRandReader(t).Read(key)
+	t.Logf("Cipher key: 0x%x", key)
+
+	block, err := mb(key)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	blockSize := block.BlockSize()
+
+	t.Run("Encryption", func(t *testing.T) {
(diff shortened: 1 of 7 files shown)


block_test.go
```diff
@@ -0,0 +1,261 @@
...
+import (
+	"bytes"
+	"crypto/cipher"
+	"io"
+	"math/rand"
+	"testing"
+	"time"
+)
+
+// This file is a copy of https://github.com/golang/go/blob/9e9b1f57c26a6d13fdaebef67136718b8042cdba/src/crypto/internal/cryptotest/block.go.
```
dagood, line 12: The modifications aren't super subtle (if someone updates this file to a new commit, I don't think they'd be confused about what they need to change), but it seems worthwhile to either mention the changes or rearrange this to make the copy purer.

(E.g. copy it as-is to `internal/cryptotest/block_test.go` so that exporting `TestBlock` still makes sense, and add a new file to fill in `newRandReader` without modifying this one.)

Diff for reference:

```diff
-// testBlock performs a set of tests on cipher.Block implementations, checking
+// TestBlock performs a set of tests on cipher.Block implementations, checking
 // the documented requirements of BlockSize, Encrypt, and Decrypt.
-func testBlock(t *testing.T, keySize int, mb MakeBlock) {
+func TestBlock(t *testing.T, keySize int, mb MakeBlock) {
 	// Generate random key
 	key := make([]byte, keySize)
 	newRandReader(t).Read(key)
@@ -239,12 +236,6 @@ func testCipher(t *testing.T, cipher func(dst, src []byte), blockSize int) {
 	})
 }
 
-func newRandReader(t *testing.T) io.Reader {
-	seed := time.Now().UnixNano()
-	t.Logf("Deterministic RNG seed: 0x%x", seed)
-	return rand.New(rand.NewSource(seed))
-}
-
 func mustPanic(t *testing.T, msg string, f func()) {
 	t.Helper()
 ```

dagood (approved):
A factoring suggestion, but functionally LGTM.

ueno (approved):

Issue #148, opened 2024-07-31
Implement encoding.BinaryAppender for all openssl-backed crypto hashes


dagood:
I believe we need to implement encoding.BinaryAppender here so that our hash works with this change coming in 1.24:

* 601776: crypto: implement encoding.BinaryAppender for all crypto hashes | https://go-review.googlesource.com/c/go/+/601776
* https://github.com/golang/go/issues/62384

(Haven't investigated much, but spotted the CL with a search/alert @qmuntal set up and wanted to file this.)

qmuntal:
Done in #170.

PR #170, opened 2024-09-08
Add implementation of encoding.BinaryAppender for hash algorithms

mertakman:
This PR resolves issue #148 by implementing the encoding.BinaryAppender interface for our hash functions. In Go 1.24, all hash algorithms are expected to implement this interface, as outlined in the following changeset: [crypto: implement encoding.BinaryAppender for all crypto hashes](https://go-review.googlesource.com/c/go/+/601776).

By adding this interface implementation to our fork, we ensure compatibility with the upcoming Go 1.24 release and future-proof our codebase against breaking changes related to the handling of binary encoding for hash algorithms.

### DIFF
--- a/hash.go
+++ b/hash.go
@@ -6,7 +6,6 @@ package openssl
 import "C"
 import (
 	"crypto"
-	"encoding/hex"
 	"errors"
 	"hash"
 	"runtime"
@@ -417,16 +416,6 @@ func (h *md5Marshal) AppendBinary(buf []byte) ([]byte, error) {
 	return append(buf, binaryData...), nil
 }
 
-func (h *md5Marshal) AppendText(buf []byte) ([]byte, error) {
-	binaryData, err := h.MarshalBinary()
-	if err != nil {
-		return nil, err
-	}
-	hexData := make([]byte, hex.EncodedLen(len(binaryData)))
-	hex.Encode(hexData, binaryData)
-	return append(buf, hexData...), nil
-}
-
 // NewSHA1 returns a new SHA1 hash.
 func NewSHA1() hash.Hash {
 	h := sha1Hash{evpHash: newEvpHash(crypto.SHA1)}
@@ -524,16 +513,6 @@ func (h *sha1Marshal) AppendBinary(buf []byte) ([]byte, error) {
 		return nil, err
 	}
 	return append(buf, binaryData...), nil
-}
-
-func (h *sha1Marshal) AppendText(buf []byte) ([]byte, error) {
-	binaryData, err := h.MarshalBinary()
-	if err != nil {
-		return nil, err
-	}
-	hexData := make([]byte, hex.EncodedLen(len(binaryData)))
-	hex.Encode(hexData, binaryData)
-	return append(buf, hexData...), nil
 }
 
 // NewSHA224 returns a new SHA224 hash.
@@ -731,26 +710,6 @@ func (h *sha256Marshal) AppendBinary(buf []byte) ([]byte, error) {
 		return nil, err
 	}
 	return append(buf, binaryData...), nil
-}
-
-func (h *sha224Marshal) AppendText(buf []byte) ([]byte, error) {
-	binaryData, err := h.MarshalBinary()
-	if err != nil {
-		return nil, err
-	}
-	hexData := make([]byte, hex.EncodedLen(len(binaryData)))
-	hex.Encode(hexData, binaryData)
-	return append(buf, hexData...), nil
-}
-
-func (h *sha256Marshal) AppendText(buf []byte) ([]byte, error) {
-	binaryData, err := h.MarshalBinary()
-	if err != nil {
-		return nil, err
-	}
-	hexData := make([]byte, hex.EncodedLen(len(binaryData)))
-	hex.Encode(hexData, binaryData)
-	return append(buf, hexData...), nil
 }
 
 // NewSHA384 returns a new SHA384 hash.
@@ -956,26 +915,6 @@ func (h *sha512Marshal) AppendBinary(buf []byte) ([]byte, error) {
 		return nil, err
 	}
 	return append(buf, binaryData...), nil
-}
-
-func (h *sha384Marshal) AppendText(buf []byte) ([]byte, error) {
-	binaryData, err := h.MarshalBinary()
-	if err != nil {
-		return nil, err
-	}
-	hexData := make([]byte, hex.EncodedLen(len(binaryData)))
-	hex.Encode(hexData, binaryData)
-	return append(buf, hexData...), nil
-}
-
-func (h *sha512Marshal) AppendText(buf []byte) ([]byte, error) {
-	binaryData, err := h.MarshalBinary()
-	if err != nil {
-		return nil, err
-	}
-	hexData := make([]byte, hex.EncodedLen(len(binaryData)))
-	hex.Encode(hexData, binaryData)
-	return append(buf, hexData...), nil
 }
 
 // NewSHA3_224 returns a new SHA3-224 hash.
--- a/hash.go
+++ b/hash.go
@@ -408,7 +408,7 @@ func (h *md5Marshal) AppendBinary(buf []byte) ([]byte, error) {
 	buf = appendUint32(buf, d.h[2])
 	buf = appendUint32(buf, d.h[3])
 	buf = append(buf, d.x[:d.nx]...)
-	buf = buf[:len(buf)+len(d.x)-int(d.nx)] // already zero
+	buf = append(buf, make([]byte, len(d.x)-int(d.nx))...) // already zero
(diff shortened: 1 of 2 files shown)


hash.go
```diff
@@ -358,6 +359,24 @@ func (h *md5Marshal) UnmarshalBinary(b []byte) error {
 	return nil
 }
 
+func (h *md5Marshal) AppendBinary(buf []byte) ([]byte, error) {
+	binaryData, err := h.MarshalBinary()
```
qmuntal, line 363: I appreciate that you intended to implement the new methods with the few modifications possible. On the other hand, implementing `AppendBinary` in terms of `MarshalBinary` doesn't achieve the sole benefit of the `encoding.BinaryAppender` interface: not allocating a new slice every time it is called.

It would be much better if the code inside `MarshalBinary` is moved to `AppendBinary` (with the necessary changes), and then `MarshalBinary` is updated to call `AppendBinary`.

hash.go
```diff
@@ -358,6 +359,24 @@ func (h *md5Marshal) UnmarshalBinary(b []byte) error {
...
+	if err != nil {
+		return nil, err
+	}
+	return append(buf, binaryData...), nil
+}
+
+func (h *md5Marshal) AppendText(buf []byte) ([]byte, error) {
```
qmuntal, line 370: The standard library hashers don't implement `encoding.AppendText`, so there is no need for us to implement that on our side. The less code we maintain, the better.

qmuntal (changes requested):
The standard library also implements the [encoding.BinaryUnmarshaler](https://pkg.go.dev/encoding#BinaryUnmarshaler) for all crypto hashers. We should do that too.

qmuntal (changes requested):
Tests missing.

hash.go
mertakman: Yes, thanks for recommending this approach 👍. I applied it and now looks much better.

hash.go
```diff
@@ -396,21 +401,7 @@ type sha1Marshal struct {
...
-	b = appendUint32(b, d.h[0])
-	b = appendUint32(b, d.h[1])
-	b = appendUint32(b, d.h[2])
-	b = appendUint32(b, d.h[3])
-	b = appendUint32(b, d.h[4])
-	b = append(b, d.x[:d.nx]...)
-	b = b[:len(b)+len(d.x)-int(d.nx)] // already zero
-	b = appendUint64(b, uint64(d.nl)>>3|uint64(d.nh)<<29)
-	return b, nil
+	return h.AppendBinary(nil)
```
qmuntal, line 404: We also know the buffer size here.

```suggestion
	buf := make([]byte, 0, sha1MarshaledSize)
	return h.AppendBinary(buf)
```

qmuntal (changes requested):

hash.go
```diff
@@ -358,6 +346,23 @@ func (h *md5Marshal) UnmarshalBinary(b []byte) error {
...
+		return nil, errors.New("crypto/md5: can't retrieve hash state")
+	}
+
+	buf = append(buf, md5Magic...)
+	buf = appendUint32(buf, d.h[0])
+	buf = appendUint32(buf, d.h[1])
+	buf = appendUint32(buf, d.h[2])
+	buf = appendUint32(buf, d.h[3])
+	buf = append(buf, d.x[:d.nx]...)
+	buf = buf[:len(buf)+len(d.x)-int(d.nx)] // already zero
```
qmuntal, line 361: This seems more robust: `b = append(b, make([]byte, len(d.x)-d.nx)...)`. Same for other marshalers.

qmuntal (changes requested):

hash.go
mertakman: fixed

hash.go
mertakman: fixed

qmuntal:
This comment is still not resolved: https://github.com/golang-fips/openssl/pull/170#pullrequestreview-2288871034

hash_test.go
```diff
@@ -78,6 +78,59 @@ func TestHash(t *testing.T) {
...
+				if err != nil {
+					t.Errorf("could not marshal: %v", err)
+				}
+				h2 := cryptoToHash(ch)()
+				if err := h2.(encoding.BinaryUnmarshaler).UnmarshalBinary(state); err != nil {
+					t.Errorf("could not unmarshal: %v", err)
+				}
+				if actual, actual2 := h.Sum(nil), h2.Sum(nil); !bytes.Equal(actual, actual2) {
+					t.Errorf("0x%x != marshaled 0x%x", actual, actual2)
+				}
```
qmuntal, line 93: This is already tested in `TestHash_BinaryMarshaler`

hash_test.go
```diff
@@ -78,6 +78,59 @@ func TestHash(t *testing.T) {
...
+				appendedState := state[10:]
+
+				h2 := cryptoToHash(ch)()
+				if err := h2.(encoding.BinaryUnmarshaler).UnmarshalBinary(appendedState); err != nil {
+					t.Errorf("could not unmarshal: %v", err)
+				}
+				if actual, actual2 := h.Sum(nil), h2.Sum(nil); !bytes.Equal(actual, actual2) {
+					t.Errorf("0x%x != appended 0x%x", actual, actual2)
+				}
+			}
```
qmuntal, line 132: Factor out this block into its own test (e.g. `TestHash_AppendBinary`.

hash_test.go
```diff
@@ -78,6 +78,59 @@ func TestHash(t *testing.T) {
...
+				for i := range prebuiltSlice {
+					prebuiltSlice[i] = byte(i)
+				}
+
+				// Clone the prebuilt slice for comparison
+				prebuiltSliceClone := append([]byte(nil), prebuiltSlice...)
+
+				// Append binary data to the prebuilt slice
+				state, err := h.(interface {
+					AppendBinary(b []byte) ([]byte, error)
```
qmuntal, line 111: You can avoid this type cast by using a named variable in here https://github.com/golang-fips/openssl/pull/170/files#diff-848945eacf9cfc5d080eda9e5c42ac8adc3afbe413745df09dbae08ae011a47aR97

qmuntal (changes requested):

mertakman:
> encoding.BinaryUnmarshaler

UnmarshalBinary method is already present at the hash file.

karianna (commented):
Given this was introduced as part of Go 1.24 do we need to document that fact in the source code somewhere?  I also assume there's no backwards compatibly breaking occurring here?

qmuntal (approved):
Other than a minor comment, looks great!

hash_test.go
```diff
@@ -117,6 +117,54 @@ func TestHash_BinaryMarshaler(t *testing.T) {
 	}
 }
 
+func TesHash_BinaryAppender(t *testing.T) {
```
dagood, line 120: This func won't quite be run as a test yet: 😄 

```suggestion
func TestHash_BinaryAppender(t *testing.T) {
```

hash.go
```diff
@@ -408,6 +396,23 @@ func (h *md5Marshal) UnmarshalBinary(b []byte) error {
...
+		return nil, errors.New("crypto/md5: can't retrieve hash state")
+	}
+
+	buf = append(buf, md5Magic...)
+	buf = appendUint32(buf, d.h[0])
+	buf = appendUint32(buf, d.h[1])
+	buf = appendUint32(buf, d.h[2])
+	buf = appendUint32(buf, d.h[3])
+	buf = append(buf, d.x[:d.nx]...)
+	buf = append(buf, make([]byte, len(d.x)-int(d.nx))...) // already zero
```
dagood, line 411: I don't think the comment is correct in the new context, at least the way I read it. This code isn't always dealing with preallocated zeros--the zeros may or may not already exist.

I don't think this needs any comment anymore. The new approach is more intuitive, "normal" Go: it's clear from the line alone that we're appending bytes of 0, whether or not they already existed.

(Same for the other `// already zero` comments being moved.)

hash_test.go
```diff
@@ -117,6 +117,54 @@ func TestHash_BinaryMarshaler(t *testing.T) {
...
+		t.Run(ch.String(), func(t *testing.T) {
+			t.Parallel()
+			if !openssl.SupportsHash(ch) {
+				t.Skip("skipping: not supported")
+			}
+			h := cryptoToHash(ch)()
+			if h, ok := h.(interface {
+				AppendBinary(b []byte) ([]byte, error)
+				Sum(b []byte) []byte
+			}); ok {
```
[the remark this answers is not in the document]
mertakman: sure,fixed

gdams (approved):

hash_test.go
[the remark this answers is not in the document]
dagood: This doesn't seem to be changed?

qmuntal:
@derekparker could you take a look at this? Thanks!

derekparker (approved):
LGTM

karianna (approved):

PR #171, opened 2024-09-09
Fix: Loop Variable Capturing Issue in Go 1.22

mertakman:
This pull request addresses an issue with loop variable capturing that was resolved in Go 1.22. Prior to Go 1.22, closures within loops would capture the loop variable itself, leading to unexpected behavior due to the variable's value changing by the time the closure was executed.

Changes Made:
Updated code to align with the new behavior in Go 1.22, where the loop variable's value is captured at each iteration rather than the variable itself.

### DIFF
--- a/openssl_test.go
+++ b/openssl_test.go
@@ -58,7 +58,7 @@ func TestMain(m *testing.M) {
 	fmt.Println("OpenSSL version:", openssl.VersionText())
 	fmt.Println("FIPS enabled:", openssl.FIPS())
 	status := m.Run()
-	for _ = range 5 {
+	for range 5 {
 		// Run GC a few times to avoid false positives in leak detection.
 		runtime.GC()
 		// Sleep a bit to let the finalizers run.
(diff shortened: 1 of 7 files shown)


qmuntal (approved):
Thanks!

qmuntal (approved):

openssl_test.go
```diff
@@ -58,7 +58,7 @@ func TestMain(m *testing.M) {
 	fmt.Println("OpenSSL version:", openssl.VersionText())
 	fmt.Println("FIPS enabled:", openssl.FIPS())
 	status := m.Run()
-	for i := 0; i < 5; i++ {
+	for _ = range 5 {
```
dagood, line 61: Is the assignment necessary?

```suggestion
	for range 5 {
```

dagood (approved):
LGTM, but to walk through a train of thought on why:

My concern with this would be if it makes it riskier to port changes back to support branches. In particular, for Linux distros that support Go for longer than Google, they may still need to support 1.21 (or earlier). If these changes are accidentally ported back, they would cause issues, so ports have to be more careful.

*However*, backports won't be safe just because we don't merge this PR. Now that the module is `go 1.22`, these dependencies can be anywhere, introduced over time, and it would be bad to assume that *any* Go code backport to a pre-1.22 version is safe. (Code that depends tightly on loop var capture behavior could cause some wild issues pre-1.22.)

So: that concern isn't valid, and cleaning up the code is good, as it normally is in ordinary codebases. (Unless we revisit earlier decisions.)

Issue #165, opened 2024-09-04
Replace capital first letter for param and return names


dagood:
nit: why CAPS for the `X` `Y` param names?

_Originally posted by @derekparker in https://github.com/golang-fips/openssl/pull/135#discussion_r1742418349_
            

There are a decent number of existing names that can be cleaned up too--a few examples:

```
type BigInt
func GenerateKeyECDSA(curve string) (X, Y, D BigInt, err error)
func GenerateKeyRSA(bits int) (N, E, D, P, Q, Dp, Dq, Qinv BigInt, err error)

type PrivateKeyECDSA
func NewPrivateKeyECDSA(curve string, X, Y, D BigInt) (*PrivateKeyECDSA, error)

type PrivateKeyRSA
func NewPrivateKeyRSA(N, E, D, P, Q, Dp, Dq, Qinv BigInt) (*PrivateKeyRSA, error)
```

I've chalked this up to "spelling it how cryptographers do" in the past, so I'm indifferent, but we should probably decide.

mertakman:
#172  solved this, please close this issue

PR #172, opened 2024-09-09
fix:Decapitalize parameters

mertakman:
In Go, it is a common convention to use lowercase letters for parameter names to maintain consistency and improve readability. 
Solves #165


qmuntal (approved):
Looks better, thanks!

PR #182, opened 2024-09-13
 Refactor RandReader Implementation

mertakman:
Changes
-Changed randReader from int to struct{}.
   -Semantic clarity: Clearly indicates the type doesn't need to hold state.
   -Consistency: Aligns with Go conventions for types that only need methods. (for example: io.Discard)
-Replaced global RandReader variable with NewRandReader() function.

### DIFF
--- a/dsa_test.go
+++ b/dsa_test.go
@@ -111,7 +111,7 @@ func testDSASignAndVerify(t *testing.T, priv *openssl.PrivateKeyDSA) {
 	if !dsa.Verify(&priv1.PublicKey, hashed[:], esig.R, esig.S) {
 		t.Error("compat: crypto/dsa can't verify OpenSSL signature")
 	}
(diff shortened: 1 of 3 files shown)


qmuntal (commented):
This is a breaking change. We would have to create a `v3` release if this PR was merged, which we don't want to do unless there are very good reasons.

mertakman:
> This is a breaking change. We would have to create a `v3` release if this PR was merged, which we don't want to do unless there are very good reasons.
> 
> The main benefit of the current definition is that it is a drop-in repalcement for the boring `RandReader`: https://github.com/golang/go/blob/8a4b439ee6e0f5d41cb6ae30146b901f3c0ad4b7/src/crypto/internal/boring/rand.go. This means less patching in our forked Go standard library.

ah okey than, nevermind.

PR #184, opened 2024-09-13
Refactor openssl.FIPS to support third-party FIPS providers

qmuntal:
Our current `openssl.FIPS` implementation only works with the built-in FIPS provider, as it checks whether it is available or not to determine if FIPS is enabled, which precludes using other FIPS providers.

This PR removes the `OSSL_PROVIDER_available(nil, "fips")` check from `openssl.FIPS` so it is not tied to the built-in provider.

The new implementation should be as good as the previous: if `EVP_default_properties_is_fips_enabled` returns `1`, we can be sure that only providers matching the `fips=1` property will be used, so we can say that FIPS mode is enabled. To avoid breacking backwards compatibility, I've added an additional check that tried to fetch a SHA256 digets using the default properties (which should contain `fips=1` at that point). If it succeeds, then it means that there is a provider that matches the `fips=1` query.

### DIFF
--- a/goopenssl.c
+++ b/goopenssl.c
@@ -35,6 +35,18 @@ FOR_ALL_OPENSSL_FUNCTIONS
 #undef DEFINEFUNC_RENAMED_1_1
 #undef DEFINEFUNC_RENAMED_3_0
 
+// go_openssl_fips_enabled returns 1 if FIPS mode is enabled, 0 otherwise.
+// See openssl.FIPS for details about its implementation.
+//
+// This function is reimplemented here because openssl.FIPS assumes that
+// all the OpenSSL bindings are loaded, that is, go_openssl_load_functions has
+// already been called. On the other hand, go_openssl_fips_enabled is called from
+// openssl.CheckFIPS, which is used to check if a given OpenSSL shared library
+// exists and is FIPS compliant. That shared library might not be the one that
+// was passed to go_openssl_load_functions, or it might not even have been called at all.
+//
+// It is written in C because it is not possible to directly call C function pointers
+// retrieved using dlsym from Go.
 int
 go_openssl_fips_enabled(void* handle)
 {
@@ -45,12 +57,8 @@ go_openssl_fips_enabled(void* handle)
         return FIPS_mode();
 
     // For OpenSSL 3.x.
-    int (*EVP_default_properties_is_fips_enabled)(void*);
-    int (*OSSL_PROVIDER_available)(void*, const char*);
-    EVP_default_properties_is_fips_enabled = (int (*)(void*))dlsym(handle, "EVP_default_properties_is_fips_enabled"); 
-    OSSL_PROVIDER_available = (int (*)(void*, const char*))dlsym(handle, "OSSL_PROVIDER_available"); 
-    if (EVP_default_properties_is_fips_enabled != NULL && OSSL_PROVIDER_available != NULL &&
-        EVP_default_properties_is_fips_enabled(NULL) == 1 && OSSL_PROVIDER_available(NULL, "fips") == 1)
+    int (*EVP_default_properties_is_fips_enabled)(void*) = (int (*)(void*))dlsym(handle, "EVP_default_properties_is_fips_enabled"); 
+    if (EVP_default_properties_is_fips_enabled != NULL && EVP_default_properties_is_fips_enabled(NULL) == 1)
             return 1;
 
     return 0;
--- a/goopenssl.c
+++ b/goopenssl.c
@@ -36,6 +36,7 @@ FOR_ALL_OPENSSL_FUNCTIONS
 #undef DEFINEFUNC_RENAMED_3_0
 
 // go_openssl_fips_enabled returns 1 if FIPS mode is enabled, 0 otherwise.
+// As a special case, it returns -1 if it cannot determine if FIPS mode is enabled.
 // See openssl.FIPS for details about its implementation.
 //
 // This function is reimplemented here because openssl.FIPS assumes that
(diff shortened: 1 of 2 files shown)


dagood (approved):
> The downside is that if FIPS mode is configured (aka `fips=1` in the default properties) but the FIPS provider is missing, then `openssl.FIPS` will return true and the user won't notice that FIPS mode is unusable until it tries to use an algorithm, e.g. calling `sha256.New`.

It looks like if a Go fork wanted to keep this init-time check working and only expects to ever run with the built in OpenSSL FIPS provider, there would just need to be a new Go-side func added to let that Go fork run the old check. (But if nobody actually cares, there's no need to add a func like that.)

qmuntal:
@dagood I made up my mind a bit with this this PR. I was too much of a breaking change. I've added a provider-agnostic check so the new behavior is less disruptive. Please take another look.

openssl.go
```diff
@@ -96,23 +96,36 @@ func VersionText() string {
...
 func FIPS() bool {
 	switch vMajor {
 	case 1:
 		return C.go_openssl_FIPS_mode() == 1
 	case 3:
-		// If FIPS is not enabled via default properties, then we are sure FIPS is not used.
-		if C.go_openssl_EVP_default_properties_is_fips_enabled(nil) == 0 {
+		// FIPS is not enabled via default properties (i.e. `fips=1`), then we are sure FIPS is not used.
+		// Note that it is still possible that the provider used by default is FIPS-compliant,
+		// but that wouldn't be a system or user requirement.
```
dagood, line 112: It sounds like this is contradictory, "we are sure FIPS is not used", yet "it is still possible that the provider used by default is FIPS-compliant". Maybe it's unclear what it means to "use" FIPS? Or should the "sure" be softened to a known working case (more like the next comment)?

goopenssl.c
```diff
@@ -57,11 +57,24 @@ go_openssl_fips_enabled(void* handle)
...
-    if (EVP_default_properties_is_fips_enabled != NULL && EVP_default_properties_is_fips_enabled(NULL) == 1)
-            return 1;
+    int (*EVP_default_properties_is_fips_enabled)(void*) = (int (*)(void*))dlsym(handle, "EVP_default_properties_is_fips_enabled");
+    void *(*EVP_MD_fetch)(void*, const char*, const char*) = (void* (*)(void*, const char*, const char*))dlsym(handle, "EVP_MD_fetch");
+    void (*EVP_MD_free)(void*) = (void (*)(void*))dlsym(handle, "EVP_MD_free");
 
-    return 0;
+    if (EVP_default_properties_is_fips_enabled == NULL || EVP_MD_fetch == NULL || EVP_MD_free == NULL) {
+        // Shouldn't happen, but if it does, we can't determine if FIPS mode is enabled.
+        return 0;
```
dagood, line 66: A false negative for FIPS mode does seem nearly harmless, but it also seems like it would be pretty simple to return and handle `-1` or `2` or something just to be sure.

goopenssl.c
```diff
@@ -35,6 +35,18 @@ FOR_ALL_OPENSSL_FUNCTIONS
 #undef DEFINEFUNC_RENAMED_1_1
 #undef DEFINEFUNC_RENAMED_3_0
 
+// go_openssl_fips_enabled returns 1 if FIPS mode is enabled, 0 otherwise.
+// See openssl.FIPS for details about its implementation.
+//
+// This function is reimplemented here because openssl.FIPS assumes that
+// all the OpenSSL bindings are loaded, that is, go_openssl_load_functions has
+// already been called. On the other hand, go_openssl_fips_enabled is called from
+// openssl.CheckFIPS, which is used to check if a given OpenSSL shared library
```
dagood, line 44:
```suggestion
// openssl.CheckVersion, which is used to check if a given OpenSSL shared library
```

goopenssl.c
qmuntal: The only thing with the current `openssl.CheckVersion` API is to log the error or panic, as it doesn't return any error value. I wouldn't say it is a false negative, if those functions are not available then we won't be able to use the shared library properly, even less in FIPS mode.

openssl.go
qmuntal: Good point. I've improved the comments a bit.

qmuntal:

goopenssl.c
dagood: The sequence of events I'm thinking about is that the user called `CheckVersion(version string) (exists, fips bool)` on a "bad" version of OpenSSL that doesn't define one of these functions, and then `CheckVersion` returns `true, false`. This implies that it exists and *isn't* in FIPS mode (given the current docs), but technically we just can't tell what mode it's in.

I guess the more realistic oddity is the `exists = true`. If the caller of `CheckVersion` is trying to decide which version to load based on this info, they might try to use the broken version that we already know we can't use properly. Maybe it would be better to return `false, false`?

In practice I don't think this is likely to ever matter, the code just makes this edge case relatively easy to spot. 😄

qmuntal: > I guess the more realistic oddity is the exists = true.

Good point. In fact, the current implementation returns `exists=true` if the shared library loaded is not OpenSSL, which is weird. I'll update the code with your recommendations. Thanks!

dagood (approved):

gdams (approved):

qmuntal:
> @derekparker could someone from your team review this PR? Thanks!

Friendly ping. @ueno @derekparker

derekparker:
@ueno I will defer to you on this one, but from my perspective it LGTM.

qmuntal:
@derekparker I'm merging this PR, as this is blocking some other work I have to do. Feel free to open an issue with any concern you have once your team have time to review it.

PR #185, opened 2024-09-13
Make algorithms more robust to unsupported hashes

qmuntal:
While testing different provider configurations I found out that we are still panicking instead of returning an error in several places where the desired hash function is not supported by the OpenSSL provider.

This PR fixes all cases where we are calling a user-provided `func() hash.Hash` function that can panic, e.g. when calling `openssl.NewHMAC(openssl.NewSHA224, nil)` with a provider that doesn't support SHA224 (aka SymCrypt).

Our azurelinux CI job hasn't triggered this situation because it is configured to use SymCrypt by default and fall back to the built-in default provider for algorithms that SymCrypt doesn't support. I triggered the error by making the built-in default provider unavailable. Anyway, I've added some tests so that we cover this situation with any provider configuration.

### DIFF
--- a/evp.go
+++ b/evp.go
@@ -16,15 +16,28 @@ import (
 // cacheMD is a cache of crypto.Hash to GO_EVP_MD_PTR.
 var cacheMD sync.Map
 
-// hashFuncHash returns fn(). If fn() panics, it returns nil.
+// hashFuncHash calls fn() and returns its result.
+// If fn() panics, the panic is recovered and returned as an error.
 // This is used to avoid aborting the program when calling
 // an unsupported hash function. It is the caller's responsibility
 // to check the returned value.
-func hashFuncHash(fn func() hash.Hash) hash.Hash {
+func hashFuncHash(fn func() hash.Hash) (h hash.Hash, err error) {
 	defer func() {
-		_ = recover()
+		r := recover()
+		if r == nil {
+			return
+		}
+		h = nil
+		switch e := r.(type) {
+		case error:
+			err = e
+		case string:
+			err = errors.New(e)
+		default:
+			err = errors.New("unsupported panic")
+		}
 	}()
-	return fn()
+	return fn(), nil
 }
 
 // hashToMD converts a hash.Hash implementation from this package to a GO_EVP_MD_PTR.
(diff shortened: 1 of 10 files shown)


evp.go
```diff
@@ -16,6 +16,17 @@ import (
 // cacheMD is a cache of crypto.Hash to GO_EVP_MD_PTR.
 var cacheMD sync.Map
 
+// hashFuncHash return fn(). If fn panics, it returns nil.
```
dagood, line 19:
```suggestion
// hashFuncHash returns fn(). If fn() panics, it returns nil.
```

evp.go
```diff
@@ -16,6 +16,17 @@ import (
...
+// This is used to avoid aborting the program when calling
+// an unsupported hash function. It is the caller's responsibility
```
dagood, line 21: Is the *only* possible panic due to lack of support? (It seems like this is the claim that would make this ok, and I think it deserves a direct mention and justification if possible.) Otherwise, preserving the error message seems like it would make sense because it would be important to figure other cases out--maybe even to know whether they're happening.

hkdf_test.go
```diff
@@ -387,3 +396,11 @@ func TestHKDFLimit(t *testing.T) {
 		t.Errorf("key expansion overflowed: n = %d, err = %v", n, err)
 	}
 }
+
+func TestHKDFUnsupportedHash(t *testing.T) {
+	// Test that newHKDF returns and error instead of panicking.
```
dagood, line 401:
```suggestion
	// Test that newHKDF returns an error instead of panicking.
```

evp.go
qmuntal: Good point. I'll preserve the error wherever possible.

dagood (approved):

qmuntal:

hkdf_test.go
```diff
@@ -314,7 +314,10 @@ func TestHKDF(t *testing.T) {
 			t.Errorf("test %d: incorrect PRK: have %v, need %v.", i, prk, tt.prk)
 		}
 
-		hkdf := newHKDF(tt.hash, tt.master, tt.salt, tt.info)
+		hkdf, err := newHKDF(tt.hash, tt.master, tt.salt, tt.info)
+		if err != nil {
+			t.Errorf("test %d: error creating HKDF: %v.", i, err)
```
ueno, line 319: Should this be `t.Fatalf` or followed by a `continue`?

ueno (approved):
LGTM

PR #186, opened 2024-09-13
Only implement the supported AES extra mode interfaces

qmuntal:
Some OpenSSL providers (such as SymCrypt) doesn't implement the AES CRT mode. If there is no fall back provider for AES CRT, then `openssl.NewAESCipher` should return an object that doesn't implement the [aes.ctrAble](https://github.com/golang/go/blob/8a4b439ee6e0f5d41cb6ae30146b901f3c0ad4b7/src/crypto/aes/modes.go#L35-L37) interface. The same applies to other AES modes.

Our azurelinux CI job hasn't triggered this situation because it is configured to use SymCrypt by default and fall back to the built-in default provider for algorithms that SymCrypt doesn't support. I triggered the error by making the built-in default provider unavailable.


aes.go
```diff
@@ -9,19 +9,11 @@ import (
...
-	NewGCMTLS() (cipher.AEAD, error)
-}
-
-var _ extraModes = (*aesCipher)(nil)
-
+// NewAESCipher creates and returns a new AES cipher.Block.
+// The key argument should be the AES key, either 16, 24, or 32 bytes to select
+// AES-128, AES-192, or AES-256.
+// The returned cipher.Block can implement the CBC, CTR, and/or GCM modes if
+// the underlying OpenSSL library supports them.
```
dagood, line 16: A bit more direct:

```suggestion
// The returned cipher.Block implements the CBC, CTR, and/or GCM modes if
// the underlying OpenSSL library supports them.
```

aes.go
```diff
@@ -38,63 +30,160 @@ func NewAESCipher(key []byte) (cipher.Block, error) {
...
+		block = cipherWithCBC_GCM{aes,
+			cipherWithCBC{aes},
+			cipherWithGCM{aes},
+		}
+	case !cbcSupportes && ctrSupportes && gcmSupportes:
+		block = cipherWithCTR_GCM{aes,
+			cipherWithCTR{aes},
+			cipherWithGCM{aes},
+		}
+	case cbcSupportes && !ctrSupportes && !gcmSupportes:
```
dagood, line 60: When this kind of thing is necessary, personally I think sticking with the bool pattern is easier to read/review. Also, `supportes` -> `supported`:

```go
	var block cipher.Block
	cbcSupported := loadCipher(kind, cipherModeCBC) != nil
	ctrSupported := loadCipher(kind, cipherModeCTR) != nil
	gcmSupported := loadCipher(kind, cipherModeGCM) != nil
	aes := aesCipher{c}
	switch {
	case cbcSupported && ctrSupported && gcmSupported:
		block = cipherWithCBC_CTR_GCM{aes,
			cipherWithCBC{aes},
			cipherWithCTR{aes},
			cipherWithGCM{aes},
		}
	case cbcSupported && ctrSupported && !gcmSupported:
		block = cipherWithCBC_CTR{aes,
			cipherWithCBC{aes},
			cipherWithCTR{aes},
		}
	case cbcSupported && !ctrSupported && gcmSupported:
		block = cipherWithCBC_GCM{aes,
			cipherWithCBC{aes},
			cipherWithGCM{aes},
		}
	case cbcSupported && !ctrSupported && !gcmSupported:
		block = cipherWithCBC{aes}
	case !cbcSupported && ctrSupported && gcmSupported:
		block = cipherWithCTR_GCM{aes,
			cipherWithCTR{aes},
			cipherWithGCM{aes},
		}
	case !cbcSupported && ctrSupported && !gcmSupported:
		block = cipherWithCTR{aes}
	case !cbcSupported && !ctrSupported && gcmSupported:
		block = cipherWithGCM{aes}
	case !cbcSupported && !ctrSupported && !gcmSupported:
		block = aes
	default:
		panic("unreachable")
	}
```

(Simpler pattern: the first half should have CBC, the second half should not; the first half of each half should have CTR, the others should not; every other one should have GCM.)

My mind goes to code generation, but it would be the first in the repo (other than vectors_test.go), and this is maybe not quite bad enough to open *that* box... 😅

aes.go
```diff
@@ -9,19 +9,11 @@ import (
 	"errors"
 )
 
-type extraModes interface {
-	// Copied out of crypto/aes/modes.go.
-	NewCBCEncrypter(iv []byte) cipher.BlockMode
-	NewCBCDecrypter(iv []byte) cipher.BlockMode
-	NewCTR(iv []byte) cipher.Stream
-	NewGCM(nonceSize, tagSize int) (cipher.AEAD, error)
```
dagood, line 17: Consider keeping some copied funcs around in interface declarations and use them to "prove" which funcs are implemented by each of the combined structs rather than just commenting which ones are implemented?

qmuntal: Done.

aes.go
qmuntal: > supportes -> supported:

Done

> (Simpler pattern: the first half should have CBC, the second half should not; the first half of each half should have CTR, the others should not; every other one should have GCM.)

Agree!

> My mind goes to code generation

It will be necessary if we ever add another mode, agree. For now it is manageable.

dagood (approved):

qmuntal:

aes.go
```diff
@@ -38,63 +30,226 @@ func NewAESCipher(key []byte) (cipher.Block, error) {
...
 		return nil, err
 	}
-	return &aesCipher{c}, nil
+	var block cipher.Block
+	cbcSupported := loadCipher(kind, cipherModeCBC) != nil
+	ctrSupported := loadCipher(kind, cipherModeCTR) != nil
+	gcmSupported := loadCipher(kind, cipherModeGCM) != nil
+	aes := aesCipher{c}
+	switch {
+	case cbcSupported && ctrSupported && gcmSupported:
```
ueno, line 39: I don't have a concrete idea, but wouldn't it be possible to somehow auto-generate this permutation? Otherwise, if we want to support more modes (e.g., XTS, which is already in x/crypto), the whole internal logic would need a change?

qmuntal: @dagood also proposed something similar. Now is 2vs1, and the XTS argument is strong. I'll autogenerate those optional AES modes 👍.

qmuntal:
@ueno can you take another look? Thanks!

qmuntal:
> @ueno can you take another look? Thanks!

ping @ueno

gdams (approved):

qmuntal:
@derekparker I'm merging this PR, as this is blocking some other work I have to do. Feel free to open an issue with any concern you have once your team have time to review it.

Issue #189, opened 2024-09-13
Implement KDF algorithms using the EVP_KDF API


qmuntal:
The [EVP_KDF](https://docs.openssl.org/master/man3/EVP_KDF/) API was added in OpenSSL 3 to properly support KDF algorithms. It is recommended that KDF operations migrates from the `EVP_PKEY` API to this new API. The built-in providers still support the `EVP_PKEY` for back compat, but third party provider might not do it (SymCrypt looking at you).

We should implement all KDF algorithms using the `EVP_KDF` API to be more compatible with third-part providers.

qmuntal:
Done.

Issue #158, opened 2024-09-03
Support the SymCrypt provider for OpenSSL


qmuntal:
This is an umbrella issue to track support for the [SymCrypt provider for OpenSSL](https://github.com/microsoft/SymCrypt-OpenSSL) (SCOSSL). SCOSSL is the the default provider in [Azure Linux 3](https://github.com/microsoft/AzureLinux), so we (the Microsoft maintainers of golang-fips/openssl) are very much interested on supporting SCOSSL as a first class target.

SCOSSL does not implement all algorithms supported by this repo, e.g. HMAC-SHA224, and also makes some assumptions that are incompatible with how we use the OpenSSL API. We should revist our assumptions and fix whatever needs fixing.

List of known issues/tasks:

- https://github.com/golang-fips/openssl/pull/154
- https://github.com/golang-fips/openssl/pull/155
- https://github.com/golang-fips/openssl/pull/159
- https://github.com/golang-fips/openssl/pull/161
- https://github.com/golang-fips/openssl/pull/163
- https://github.com/golang-fips/openssl/pull/167
- https://github.com/golang-fips/openssl/pull/175
- https://github.com/golang-fips/openssl/pull/178
- https://github.com/golang-fips/openssl/pull/184
- https://github.com/golang-fips/openssl/pull/185
- https://github.com/golang-fips/openssl/pull/186
- https://github.com/golang-fips/openssl/pull/187
- https://github.com/golang-fips/openssl/pull/188
- https://github.com/golang-fips/openssl/pull/191
- https://github.com/golang-fips/openssl/pull/194
- https://github.com/golang-fips/openssl/pull/206
- https://github.com/golang-fips/openssl/pull/207
- https://github.com/golang-fips/openssl/pull/211

qmuntal:
Done!

PR #191, opened 2024-09-16
Implement TLS1 PRF using the EVP_KDF API in OpenSSL 3

qmuntal:
The [EVP_KDF](https://docs.openssl.org/master/man3/EVP_KDF/) API was added in OpenSSL 3 to properly support KDF algorithms, such as TLS1 PRF. It is recommended that KDF operations migrates from the EVP_PKEY API to this new API. The built-in providers still support the EVP_PKEY for back compat, but third party provider might not do it (SymCrypt looking at you).

This PR implements TLS1 PRF using the EVP_KDF API when using OpenSSL 3.

Also, I'm introducing a new helper, `paramBuilder`,  to build parameters in a much cleaner and safer way. I'll submit a follow-up PR using this helper in other algorithms.

For #158.
For #189.

### DIFF
--- a/params.go
+++ b/params.go
@@ -73,9 +73,9 @@ func (b *paramBuilder) build() (C.GO_OSSL_PARAM_PTR, error) {
 	return param, nil
 }
 
-// addUtf8String adds a UTF-8 string to the builder.
+// addUTF8String adds a UTF-8 string to the builder.
 // size should not include the terminating NUL byte. If it is zero then it will be calculated.
-func (b *paramBuilder) addUtf8String(name *C.char, value *C.char, size C.size_t) {
+func (b *paramBuilder) addUTF8String(name *C.char, value *C.char, size C.size_t) {
 	if !b.check() {
 		return
 	}
--- a/params.go
+++ b/params.go
@@ -1,0 +1,99 @@
+//go:build !cmd_go_bootstrap
+
+package openssl
+
+// #include "goopenssl.h"
+import "C"
+import (
+	"runtime"
+	"unsafe"
+)
+
+var (
+	// KDF parameters
+	_OSSL_KDF_PARAM_DIGEST = C.CString("digest")
+	_OSSL_KDF_PARAM_SECRET = C.CString("secret")
+	_OSSL_KDF_PARAM_SEED   = C.CString("seed")
+)
+
+// paramBuilder is a helper for building OSSL_PARAMs.
+// If an error occurs when adding a new parameter,
+// subsequent calls to add parameters are ignored
+// and build() will return the error.
+type paramBuilder struct {
+	bld    C.GO_OSSL_PARAM_BLD_PTR
+	pinner runtime.Pinner
+
+	err error
+}
+
+// newParamBuilder creates a new paramBuilder.
+func newParamBuilder() (*paramBuilder, error) {
+	bld := C.go_openssl_OSSL_PARAM_BLD_new()
+	if bld == nil {
+		return nil, newOpenSSLError("OSSL_PARAM_BLD_new")
+	}
+	pb := &paramBuilder{bld: bld}
+	runtime.SetFinalizer(pb, (*paramBuilder).finalize)
+	return pb, nil
+}
+
+// free frees the builder.
+func (b *paramBuilder) finalize() {
+	if b.bld != nil {
+		b.pinner.Unpin()
+		C.go_openssl_OSSL_PARAM_BLD_free(b.bld)
+		b.bld = nil
+	}
+}
+
+func (b *paramBuilder) check() bool {
+	if b.err != nil {
+		return false
+	}
+	if b.bld == nil {
+		panic("openssl: paramBuilder has been freed")
+	}
+	return true
+}
+
+// build creates an OSSL_PARAM from the builder.
+// The returned OSSL_PARAM must be freed with OSSL_PARAM_free.
+// If an error occurred during building, it is returned.
+// The builder is freed and cannot be reused.
+func (b *paramBuilder) build() (C.GO_OSSL_PARAM_PTR, error) {
+	if !b.check() {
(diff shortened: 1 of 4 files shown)


params.go
```diff
@@ -0,0 +1,99 @@
...
+// The builder is freed and cannot be reused.
+func (b *paramBuilder) build() (C.GO_OSSL_PARAM_PTR, error) {
+	if !b.check() {
+		return nil, b.err
+	}
+	param := C.go_openssl_OSSL_PARAM_BLD_to_param(b.bld)
+	if param == nil {
+		return nil, newOpenSSLError("OSSL_PARAM_BLD_build")
+	}
+	b.finalize() // we shouldn't reuse the builder once we've built the params
```
mertakman, line 72: Why not deferring the finalizer? what if param is nil, and function returns early?

mertakman, line 72: why not calling finalizer if function fails? 
```go
func (b *paramBuilder) build() (param C.GO_OSSL_PARAM_PTR, err error) {
    if !b.check() {
        return nil, b.err
    }
    
    defer b.finalize() // Always finalize, even if an error occurs
    
    param = C.go_openssl_OSSL_PARAM_BLD_to_param(b.bld)
    if param == nil {
        err = newOpenSSLError("OSSL_PARAM_BLD_build")
    }
    
    return
}
```

qmuntal: I like the idea. Will do that, thanks!

qmuntal:
Review needed. Specially from @ueno. Thanks!

ueno (approved):
LGTM, except one minor nit

tls1prf.go
```diff
@@ -8,12 +8,20 @@ import (
...
+	"sync"
 	"unsafe"
 )
 
 func SupportsTLS1PRF() bool {
-	return vMajor > 1 ||
-		(vMajor >= 1 && vMinor >= 1)
+	switch vMajor {
+	case 1:
+		return vMajor > 1 || (vMajor >= 1 && vMinor >= 1)
```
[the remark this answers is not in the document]
qmuntal: Good catch.

tls1prf.go
```diff
@@ -102,3 +105,52 @@ func TLS1PRF(result, secret, label, seed []byte, h func() hash.Hash) error {
 	}
 	return nil
 }
+
+// fetchTLS1PRF fetches the TLS1-PRF KDF algorithm.
+// It is safe to call this function concurrently.
+// The returned EVP_KDF_PTR shouldn't be freed.
+var fetchTLS1PRF = sync.OnceValues(func() (C.GO_EVP_KDF_PTR, error) {
```
dagood, line 112: Should this be called `fetchTLS1PRF3` and/or should the comment mention OpenSSL 3?

params.go
```diff
@@ -0,0 +1,101 @@
...
+	bld := C.go_openssl_OSSL_PARAM_BLD_new()
+	if bld == nil {
+		return nil, newOpenSSLError("OSSL_PARAM_BLD_new")
+	}
+	pb := &paramBuilder{bld: bld}
+	runtime.SetFinalizer(pb, (*paramBuilder).finalize)
+	return pb, nil
+}
+
+// free frees the builder.
```
dagood, line 41:
```suggestion
// finalize frees the builder.
```

params.go
```diff
@@ -0,0 +1,101 @@
...
+func (b *paramBuilder) finalize() {
+	if b.bld != nil {
+		b.pinner.Unpin()
+		C.go_openssl_OSSL_PARAM_BLD_free(b.bld)
+		b.bld = nil
+	}
+}
+
+func (b *paramBuilder) check() bool {
```
dagood, line 50: Add doc?

```suggestion
// check is used internally to enforce invariants and should not be called by users of paramBuilder.
// Returns true if it's ok to add parameters to the builder or build it.
// Returns false if there has been an error while adding a parameter.
// Panics if the paramBuilder has been freed, e.g. if it has already been built.
func (b *paramBuilder) check() bool {
```

params.go
```diff
@@ -0,0 +1,101 @@
...
+	param := C.go_openssl_OSSL_PARAM_BLD_to_param(b.bld)
+	if param == nil {
+		return nil, newOpenSSLError("OSSL_PARAM_BLD_build")
+	}
+	b.finalize() // we shouldn't reuse the builder once we've built the params
+	return param, nil
+}
+
+// addUTF8String adds a UTF-8 string to the builder.
+// size should not include the terminating NUL byte. If it is zero then it will be calculated.
```
dagood, line 79: I'm not following the its... and who does the calculating?

shims.h
```diff
@@ -106,6 +106,9 @@ typedef void* GO_OSSL_PARAM_PTR;
 typedef void* GO_CRYPTO_THREADID_PTR;
 typedef void* GO_EVP_SIGNATURE_PTR;
 typedef void* GO_DSA_PTR;
+typedef void* GO_EVP_KDF_PTR;
+typedef void* GO_EVP_KDF_CTX_PTR;
+
```
dagood, line 111:
```suggestion
```

params.go
```diff
@@ -0,0 +1,101 @@
...
+func (b *paramBuilder) build() (C.GO_OSSL_PARAM_PTR, error) {
+	defer b.finalize()
+	if !b.check() {
+		return nil, b.err
+	}
+	param := C.go_openssl_OSSL_PARAM_BLD_to_param(b.bld)
+	if param == nil {
+		return nil, newOpenSSLError("OSSL_PARAM_BLD_build")
+	}
+	b.finalize() // we shouldn't reuse the builder once we've built the params
```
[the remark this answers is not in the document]
qmuntal: Wops, forgot to remove it in a previous refactor.

params.go
qmuntal: Reworded 👍

PR #192, opened 2024-09-17
Remove Redundant Slice Syntax

mertakman:
Simplify Slice Passing by Eliminating [:] Redundancy, and couple error formatting fix in those files.

### DIFF
--- a/ed25519.go
+++ b/ed25519.go
@@ -6,6 +6,7 @@ package openssl
 import "C"
 import (
 	"errors"
+	"fmt"
 	"runtime"
 	"strconv"
 	"sync"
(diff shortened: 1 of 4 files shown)


ed25519.go
```diff
@@ -150,7 +151,7 @@ func extractPKEYPubEd25519(pkey C.GO_EVP_PKEY_PTR, pub []byte) error {
 		return newOpenSSLError("EVP_PKEY_get_raw_public_key")
 	}
 	if r.len != publicKeySizeEd25519 {
-		return errors.New("ed25519: bad public key length: " + strconv.Itoa(int(r.len)))
+		return fmt.Errorf("ed25519: bad public key length: %d", int(r.len))
```
qmuntal, line 154: We can't depend on `fmt` here, this library is too low level for that 😛.

qmuntal (changes requested):

qmuntal (approved):

PR #195, opened 2024-09-26
Fix `SupportsHash` to correctly detect unsupported hashed on OpenSSL 1

mertakman:


evp.go
```diff
@@ -90,21 +99,13 @@ func cryptoHashToMD(ch crypto.Hash) (md C.GO_EVP_MD_PTR) {
...
 	case crypto.SHA3_384:
-		if versionAtOrAbove(1, 1, 1) {
-			return C.go_openssl_EVP_sha3_384()
-		}
+		return C.go_openssl_EVP_sha3_384()
 	case crypto.SHA3_512:
-		if versionAtOrAbove(1, 1, 1) {
-			return C.go_openssl_EVP_sha3_512()
-		}
+		return C.go_openssl_EVP_sha3_512()
```
qmuntal, line 108: CI is complaining when using OpenSSL 1.0.2 and 1.1.1 because this functions are only defined in OpenSSL 1.1.1 and above. The `versionAtOrAbove` is necessary here.

PR #196, opened 2024-09-26
Add Mariner 2 to Github Workflow

mertakman:
Our clients using Mariner 2 have been experiencing panics. To mitigate these issues, this pull request introduces changes to the GitHub workflow YAML file, enabling automated testing on Mariner 2 as well.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -89,3 +89,11 @@ jobs:
       uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
     - name: Run Test
       run: go test -v ./...
+  mariner2:
+    runs-on: ubuntu-latest
+    container: mcr.microsoft.com/oss/go/microsoft/golang:1.23.1-3-cbl-mariner2.0
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -89,3 +89,12 @@ jobs:
       uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
     - name: Run Test
       run: go test -v ./...
+  mariner2:
+    runs-on: ubuntu-latest
+    continue-on-error: true
```
qmuntal, line 94: Add a TODO comment explaining why we continue on error here and saying that this line should be removed once mariner2 is fully supported.

qmuntal (changes requested):
Thanks! Just one minor comment.

.github/workflows/test.yml
qmuntal, line 94: This property should be moved to the `Run Test` step.

qmuntal (changes requested):

qmuntal (approved):

PR #200, opened 2024-10-02
add .DS_Store to gitignore file

gdams:
I noticed in a recent PR that DS_Store files are popping up for macOS users: https://github.com/golang-fips/openssl/pull/199/files

These should be ignored from the codebase


qmuntal (approved):

mertakman (approved):

dagood:
I don't mind having this, but I think it's worth mentioning that macOS users should probably set up their [machine-wide gitignore](https://git-scm.com/docs/gitignore) to prevent this file from also showing up elsewhere. (Some repo owners do mind having extraneous .gitignore lines checked in.)

(To be clear, I'm not saying this should be mentioned in the .gitignore file, I just wanted to write this comment. 😄)

PR #202, opened 2024-10-08
Prevent RandReader.Read argument from escaping to the heap

qmuntal:
Go 1.24 will bring back `noescape` and `nocallback` cgo directives, this time without the issue that made us revert the previous attempt to use them (see https://github.com/golang-fips/openssl/pull/129).

For now just use the cgo directives for `go_openssl_RAND_bytes`, as there is no other way to avoid the allocation.

### DIFF
--- a/cgo_go124.go
+++ b/cgo_go124.go
@@ -1,6 +1,16 @@
 //go:build go1.24 && !cmd_go_bootstrap
 
 package openssl
+// The following noescape and nocallback directives are used to prevent the Go
+// compiler from allocating function parameters on the heap. See
+// https://github.com/golang/go/blob/0733682e5ff4cd294f5eccb31cbe87a543147bc6/src/cmd/cgo/doc.go#L439-L461
+//
+// If possible, write a C wrapper function to optimize a call rather than using
+// this feature so the optimization will work for all supported Go versions.
+//
+// This is just a performance optimization. Only add functions that have been
+// observed to benefit from these directives, not every function that is merely
+// expected to meet the noescape/nocallback criteria.
 
 // #cgo noescape go_openssl_RAND_bytes
 // #cgo nocallback go_openssl_RAND_bytes
--- a/cgo_go124.go
+++ b/cgo_go124.go
@@ -1,0 +1,7 @@
+//go:build go1.24 && !cmd_go_bootstrap
+
+package openssl
+
+// #cgo noescape go_openssl_RAND_bytes
+// #cgo nocallback go_openssl_RAND_bytes
+import "C"
--- a/cgo_go124.go
+++ b/cgo_go124.go
@@ -1,6 +1,7 @@
 //go:build go1.24 && !cmd_go_bootstrap
 
 package openssl
+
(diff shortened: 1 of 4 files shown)


cgo_go124.go
```diff
@@ -0,0 +1,7 @@
+//go:build go1.24 && !cmd_go_bootstrap
+
+package openssl
+
```
dagood, line 4: I tried updating the comment from the earlier commit:

```suggestion
// The following noescape and nocallback directives are used to prevent the Go
// compiler from allocating function parameters on the heap. See
// https://github.com/golang/go/blob/0733682e5ff4cd294f5eccb31cbe87a543147bc6/src/cmd/cgo/doc.go#L439-L461
//
// If possible, write a C wrapper function to optimize a call rather than using
// this feature so the optimization will work for all supported Go versions.
//
// This is just a performance optimization. Only add functions that have been
// observed to benefit from these directives, not every function that is merely
// expected to meet the noescape/nocallback criteria.

```

rand_test.go
```diff
@@ -12,3 +15,19 @@ func TestRand(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestAllocations(t *testing.T) {
+	n := int(testing.AllocsPerRun(10, func() {
+		buf := make([]byte, 32)
+		openssl.RandReader.Read(buf)
+		sink ^= buf[0]
```
dagood, line 20: Seems a little odd to use the definition from pbkdf2_test.go... maybe move `sink` to openssl_test.go and explicitly mention that it's for everyone to use? (Most general test file I saw.)

rand_test.go
```diff
@@ -12,3 +15,19 @@ func TestRand(t *testing.T) {
...
+	}))
+	want := 1
+	ver := strings.TrimPrefix(runtime.Version(), "devel ")
+	if version.Compare(ver, "go1.24") >= 0 {
+		want = 0
+	}
```
dagood: I anticipate we'll want to bring this out to a utility file if we add more of these in the future. Not sure it's worth doing it proactively, though, if we don't expect more.

Either way, might be worth having a comment pointing at cgo_go124.go to explain what this is all about.

dagood (approved):
LGTM, just comments about factoring and comments.

rand_test.go
qmuntal: Good point. Done.

rand_test.go
qmuntal: Will do both, thanks!

mertakman (approved):

cgo_go124.go
```diff
@@ -0,0 +1,17 @@
+//go:build go1.24 && !cmd_go_bootstrap
+
+package openssl
+// The following noescape and nocallback directives are used to prevent the Go
```
dagood: Oops, I didn't quite get the suggestion right. Go format inserts this newline:

```suggestion
package openssl

// The following noescape and nocallback directives are used to prevent the Go
```

dagood (approved):

PR #204, opened 2024-10-08
Check proper initialization for SupportEd25519

mertakman:
This PR checks if Ed25519 is properly initialized in OpenSSL 1. And adds an extra clause for SupportEd25519 to see if initialized structure is nil or not.

### DIFF
--- a/ed25519.go
+++ b/ed25519.go
@@ -37,7 +37,7 @@ func SupportsEd25519() bool {
 	onceSupportsEd25519.Do(func() {
 		switch vMajor {
 		case 1:
-			supportsEd25519 = versionAtOrAbove(1, 1, 1) || !FIPS()
+			supportsEd25519 = versionAtOrAbove(1, 1, 1) || FIPS()
 		case 3:
 			name := C.CString("ED25519")
 			defer C.free(unsafe.Pointer(name))
--- a/ed25519.go
+++ b/ed25519.go
@@ -37,7 +37,7 @@ func SupportsEd25519() bool {
 	onceSupportsEd25519.Do(func() {
 		switch vMajor {
 		case 1:
-			supportsEd25519 = versionAtOrAbove(1, 1, 1)
+			supportsEd25519 = versionAtOrAbove(1, 1, 1) || !FIPS()
 		case 3:
 			name := C.CString("ED25519")
 			defer C.free(unsafe.Pointer(name))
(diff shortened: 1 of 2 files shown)


ed25519.go
```diff
@@ -37,7 +37,12 @@ func SupportsEd25519() bool {
 	onceSupportsEd25519.Do(func() {
 		switch vMajor {
 		case 1:
-			supportsEd25519 = versionAtOrAbove(1, 1, 1)
+			ctx := C.go_openssl_EVP_PKEY_CTX_new_id(C.GO_EVP_PKEY_ED25519, nil)
```
qmuntal: No need to do this call if `versionAtOrAbove(1, 1, 1)` is false.

qmuntal (changes requested):

ed25519.go
mertakman: In the below there is already and condition for that, ofcourse we can stop making this call too , but it is an improvement with almost no cost and adding a excessive code complexity.

mertakman:
```go
  if versionAtOrAbove(1, 1, 1) {
    ctx := C.go_openssl_EVP_PKEY_CTX_new_id(C.GO_EVP_PKEY_ED25519, nil)
    if ctx != nil {
     defer C.go_openssl_EVP_PKEY_CTX_free(ctx)
    }

    supportsEd25519 = ctx != nil
   }
```
IMO it is not looking any better than

```go
   ctx := C.go_openssl_EVP_PKEY_CTX_new_id(C.GO_EVP_PKEY_ED25519, nil)
   if ctx != nil {
    defer C.go_openssl_EVP_PKEY_CTX_free(ctx)
   }

   supportsEd25519 = versionAtOrAbove(1, 1, 1) && ctx != nil
   ```

qmuntal: But we if we short-circuit we won't be unnecessary calling `go_openssl_EVP_PKEY_CTX_new_id`.

mertakman: since its wrapped with sync.Once, it doesn't bring any detectable costs anyway, probably few nanosecond at the very beginning of the app. But if you think its important , I'm adding that change than.

mertakman: done

qmuntal (commented):
LGTM!

qmuntal (approved):

qmuntal:
I see that all mariner 2.0 tests are passing. IMO we should remove the `continue-on-error: true` from the GitHub Actions mariner job in this same PR.

gdams (approved):

dagood (approved):

PR #206, opened 2024-10-11
Refactor SetFIPS to support third-party FIPS providers

qmuntal:
Our current `openssl.SetFIPS` implementation only works with the built-in FIPS provide.

This PR still tries to load the built-in FIPS provider as a best effort, but it only fails if there is no provider that supports the `-fips=yes` query.

### DIFF
--- a/openssl.go
+++ b/openssl.go
@@ -169,20 +169,23 @@ func SetFIPS(enabled bool) error {
 		} else {
 			provName = providerNameDefault
 		}
-		// Check if there is any provider that matches props.
-		if C.go_openssl_OSSL_PROVIDER_available(nil, provName) != 1 {
-			// If not, fallback to provName provider.
-			if C.go_openssl_OSSL_PROVIDER_load(nil, provName) == nil {
-				return newOpenSSLError("OSSL_PROVIDER_try_load")
-			}
-			// Make sure we now have a provider available.
-			if C.go_openssl_OSSL_PROVIDER_available(nil, provName) != 1 {
-				return fail("SetFIPS(" + strconv.FormatBool(enabled) + ") not supported")
-			}
-		}
+		// Try to load the provider, but don't fail if it's not loaded.
+		// The built-in provides might not be present in the system, e.g. because
+		// third-party providers are being used or because the system is not well-configured.
+		C.go_openssl_OSSL_PROVIDER_try_load(nil, provName, 1)
+		C.go_openssl_ERR_clear_error()
+
+		// Enable FIPS mode in the default properties.
 		if C.go_openssl_EVP_default_properties_enable_fips(nil, mode) != 1 {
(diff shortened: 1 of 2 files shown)


karianna (approved):

openssl.go
```diff
@@ -169,20 +170,28 @@ func SetFIPS(enabled bool) error {
...
-			if C.go_openssl_OSSL_PROVIDER_load(nil, provName) == nil {
-				return newOpenSSLError("OSSL_PROVIDER_try_load")
-			}
-			// Make sure we now have a provider available.
-			if C.go_openssl_OSSL_PROVIDER_available(nil, provName) != 1 {
-				return fail("SetFIPS(" + strconv.FormatBool(enabled) + ") not supported")
-			}
+		// Try to load the provider, but don't fail if it's not loaded.
+		// The built-in provides might not be present in the system, e.g. because
+		// third-party providers are being used or because the system is not well-configured.
```
dagood: I think it's worth connecting this more explicitly to what's ok or not (also, `provides` typo):

```suggestion
		// The built-in provider might not be present in the system.
		// We don't need the built-in provider if third-party providers are being used: they are already loaded.
		// If the system is not well-configured and has no FIPS capability, this will be detected by the next steps.
```

Also, I assume from this code that we do need to explicitly try to load the built-in FIPS provider, but don't need to try to load third-party providers? Or is it also possible that someone sets up a third-party provider that provides fips=yes but does require being try-loaded explicitly? (The behavior might still be right in that case, but the comment wouldn't be. 😄)

openssl.go
```diff
@@ -169,20 +170,28 @@ func SetFIPS(enabled bool) error {
...
+			return errors.New("openssl: FIPS mode not supported by any provider")
 		}
+
+		// Enable FIPS mode in the default properties.
 		if C.go_openssl_EVP_default_properties_enable_fips(nil, mode) != 1 {
-			return newOpenSSLError("openssl: EVP_default_properties_enable_fips")
+			return newOpenSSLError("EVP_default_properties_enable_fips")
 		}
+
+		C.go_openssl_EVP_MD_free(md)
```
dagood: Missing from error case?

openssl.go
qmuntal: > Or is it also possible that someone sets up a third-party provider that provides fips=yes but does require being try-loaded explicitly? 

If someone (developer of distro maintainer) sets up a third-party provider, then they know what they are doing and will probably make the provider available by default via configuration. When using the build-in FIPS provider it is more likely that a non-OpenSSL expert is trying to run OpenSSL in FIPS mode, probably for testing purposes. So we help them by trying to load the FIPS provider if not available.

openssl.go
qmuntal: Good catch, fixed.

dagood (approved):

PR #208, opened 2024-10-29
Use paramBuilder everywhere

qmuntal:
Building parameters using the OpenSSL C API is tedious and error-prone. Better use the `paramBuilder` abstraction, which knows how and when to free memory and doesn't miss any error check.

While here, move all parameter definitions to `params.go` and make sure they are not part of the public API by prepending them with an underscore.

### DIFF
--- a/params.go
+++ b/params.go
@@ -175,7 +175,7 @@ func (b *paramBuilder) addBin(name *C.char, value []byte, private bool) {
 	if !b.check() {
 		return
 	}
-	if value == nil {
+	if len(value) == 0 {
 		// Nothing to do.
 		return
 	}
@@ -196,7 +196,7 @@ func (b *paramBuilder) addBigInt(name *C.char, value BigInt, private bool) {
 	if !b.check() {
 		return
 	}
-	if value == nil {
+	if len(value) == 0 {
 		// Nothing to do.
 		return
 	}
--- a/params.go
+++ b/params.go
@@ -175,6 +175,10 @@ func (b *paramBuilder) addBin(name *C.char, value []byte, private bool) {
 	if !b.check() {
 		return
 	}
+	if value == nil {
+		// Nothing to do.
+		return
+	}
 	bn := C.go_openssl_BN_bin2bn(base(value), C.int(len(value)), nil)
 	if bn == nil {
 		b.err = newOpenSSLError("BN_bin2bn")
@@ -190,6 +194,10 @@ func (b *paramBuilder) addBin(name *C.char, value []byte, private bool) {
 // otherwise it will be freed with BN_free.
 func (b *paramBuilder) addBigInt(name *C.char, value BigInt, private bool) {
 	if !b.check() {
+		return
+	}
+	if value == nil {
+		// Nothing to do.
 		return
 	}
 	bn := bigToBN(value)
--- a/params.go
+++ b/params.go
@@ -157,7 +157,7 @@ func (b *paramBuilder) addInt32(name *C.char, value int32) {
 	}
 }
 
-// addBN adds an GO_BIGNUM_PTR to the builder.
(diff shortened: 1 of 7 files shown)


derekparker (approved):
LGTM

params.go
```diff
@@ -117,3 +156,47 @@ func (b *paramBuilder) addInt32(name *C.char, value int32) {
...
+	if bn == nil {
+		b.err = newOpenSSLError("BN_bin2bn")
+		return
+	}
+	b.bnToFree = append(b.bnToFree, bnParam{bn, private})
+	b.addBN(name, bn)
+}
+
+// addBigInt adds a BigInt to the builder.
+// The BigInt is converted to a BIGNUM using bigToBN and freed when the builder is finalized.
```
dagood: Slightly more directly saying the BIGNUM is freed rather than the BigInt:

```suggestion
// The BigInt is converted using bigToBN to a BIGNUM that is freed when the builder is finalized.
```

params.go
```diff
@@ -117,3 +156,47 @@ func (b *paramBuilder) addInt32(name *C.char, value int32) {
 		b.err = newOpenSSLError("OSSL_PARAM_BLD_push_int32(" + C.GoString(name) + ")")
 	}
 }
+
+// addBN adds an GO_BIGNUM_PTR to the builder.
```
dagood:
```suggestion
// addBN adds a GO_BIGNUM_PTR to the builder.
```

rsa.go
```diff
@@ -377,22 +366,16 @@ func rsaSetCRTParams(key C.GO_RSA_PTR, dmp1, dmq1, iqmp BigInt) bool {
...
+		name    *C.char
+		num     BigInt
+		private bool
 	}
 
 	comps := make([]bigIntParam, 0, 8)
 
 	required := [...]bigIntParam{
-		{OSSL_PKEY_PARAM_RSA_N, n}, {OSSL_PKEY_PARAM_RSA_E, e}, {OSSL_PKEY_PARAM_RSA_D, d},
+		{_OSSL_PKEY_PARAM_RSA_N, n, false}, {_OSSL_PKEY_PARAM_RSA_E, e, false}, {_OSSL_PKEY_PARAM_RSA_D, d, false},
```
dagood: It seems like it might be possible to directly call `bld.addBigInt` for each of these rather than temporarily storing them in a slice... maybe worth adding a comment for the slice to point out why this is better if I'm missing something. 😄

dagood (approved):

rsa.go
qmuntal: Good point. There is no need to use the temporary slice. I'll add the big ints directory to the builder :).

params.go
```diff
@@ -117,3 +156,55 @@ func (b *paramBuilder) addInt32(name *C.char, value int32) {
...
+
+// addBin adds a byte slice to the builder.
+// The slice is converted to a BIGNUM using BN_bin2bn and freed when the builder is finalized.
+// If private is true, the BIGNUM will be cleared with BN_clear_free,
+// otherwise it will be freed with BN_free.
+func (b *paramBuilder) addBin(name *C.char, value []byte, private bool) {
+	if !b.check() {
+		return
+	}
+	if value == nil {
```
dagood: When I see this, I have to wonder if distinguishing nil vs. empty slices could bite us later in subtle ways... e.g. if an OpenSSL param behaves differently if passed an empty array vs. not being passed at all. (An unexpected default value?)

I suppose the alternatives are:
Should we actually pass a reference to a zero-length array if `value` is nil, and make the caller responsible for skipping the call if it determines it's possible?
Or: if it's always ok to not pass a nil/0-length param, should this just check `if len(value) == 0`?

qmuntal: Good point. Go best practices says not to distinguish between an empty and a nil slice unless necessary. I'll do that here.

PR #210, opened 2024-10-31
Add cgolessbuild test job

dagood:
* To check for regressions like https://github.com/golang-fips/openssl/pull/209

I'm not totally positive why our fork depends on this. I would have expected that there are no `import`s if cgo isn't enabled because we wouldn't try to use the openssl backend. However, I don't think it would be bad to keep this working even if we figure out and remove the dependency.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -98,9 +98,18 @@ jobs:
     - name: Run Test
       run: go test -v ./...
 
-  # The module isn't useful without cgo, but it still needs to successfully build. Broad cgo-less
-  # builds should succeed if they import this module but don't refer to anything requiring cgo. This
-  # may be necessary to build a Go toolset fork, depending on how it's implemented.
+  # Verify that golang-fips/openssl builds successfully without cgo enabled.
+  #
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -97,3 +97,14 @@ jobs:
       uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
     - name: Run Test
       run: go test -v ./...
+
+  # The module isn't useful without cgo, but it still needs to successfully build. Broad cgo-less
```
karianna: I'll confess, none of this made sense to me :-).  I think it might be a passive vs active voice thing?

gdams (approved):

qmuntal (approved):
LGTM. Can be merged once @karianna comment is resolved.

.github/workflows/test.yml
dagood: The only passive voice I see (in the way I read it) is "how it's implemented"--I do think it's rushed and scattered, though, and it shouldn't be hard to fix. 😄

karianna (approved):

PR #211, opened 2024-11-04
Implement HasFIPSProvider

qmuntal:
`FIPS()` will return false if the `fips=yes` property is not set in the default properties. This is working as intended, but it is not sufficient to determine if the provider used by default is FIPS-capable. Some distros (e.g. Azure Linux 3, see https://github.com/microsoft/azurelinux/issues/10433) don't set the `-fips=yes` property when operating in FIPS mode, but instead set a FIPS-capable provider (e.g. SymCrypt) as default provider by setting the `provider` property in the default properties (e.g. `?provider=symcryptprovider`).

This PR adds a new function, `HasFIPSProvider()`, which checks if the provider used by the default matches the `fips=yes` query.

### DIFF
--- a/openssl.go
+++ b/openssl.go
@@ -141,12 +141,12 @@ func FIPS() bool {
 
 // FIPSProvider returns true if the provider used by the default matches the `fips=yes` query.
 // Note that this function can return true even if [FIPS] returns false, because [FIPS] checks
-// whether the default properties contain `fips=1`.
-// It will always return true for OpenSSL 3 if [FIPS] returns true.
-// It will always returns false for OpenSSL 1.
+// whether the default properties contain `fips=1`. It will always return true for OpenSSL 3 if
+// [FIPS] returns true.
+// When using OpenSSL 1, this function always returns the same value as [FIPS].
 func FIPSProvider() bool {
 	if vMajor == 1 {
-		return false
+		return FIPS()
 	}
 	provFn := func(props *C.char) C.GO_OSSL_PROVIDER_PTR {
 		md := C.go_openssl_EVP_MD_fetch(nil, algorithmSHA256, props)
--- a/openssl.go
+++ b/openssl.go
@@ -141,7 +141,7 @@ func FIPS() bool {
 
 // HasFIPSProvider returns true if the provider used by the default matches the `fips=yes` query.
 // Note that this function can return true even if [FIPS] returns false, because [FIPS] checks
-// whether the default properties contain `fips=1`. It will always return true for OpenSSL 3 if
+// whether the default properties contain `fips=yes`. It will always return true for OpenSSL 3 if
 // [FIPS] returns true.
 // When using OpenSSL 1, this function always returns the same value as [FIPS].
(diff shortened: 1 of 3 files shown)


openssl.go
```diff
@@ -138,6 +139,34 @@ func FIPS() bool {
...
+	provFn := func(props *C.char) C.GO_OSSL_PROVIDER_PTR {
+		md := C.go_openssl_EVP_MD_fetch(nil, algorithmSHA256, props)
+		if md == nil {
+			C.go_openssl_ERR_clear_error()
+			return nil
+		}
+		C.go_openssl_EVP_MD_free(md)
+		return C.go_openssl_EVP_MD_get0_provider(md)
+	}
+	// Load the provider with and without the `fips=yes` query.
```
mertakman: Comment above says fips=yes and comment in function mentions fips=1, would be good to update this to fips=1 in everywhere for consistency

openssl_test.go
```diff
@@ -91,3 +92,24 @@ func compareCurrentVersion(v string) int {
...
+	if openssl.FIPS() {
+		if !fipsProv {
+			t.Fatal("FIPSProvider should be true when FIPS is enabled")
+		}
+	} else if openssl.SymCryptProviderAvailable() {
+		if !fipsProv {
+			t.Fatal("FIPSProvider should be true when SymCrypt is available")
+		}
+	}
+}
```
mertakman, line 106:
```go
func TestFIPSProvider(t *testing.T) {
	fipsProv := openssl.FIPSProvider()

	switch {
	case openssl.MajorVersion() == 1:
		if fipsProv != openssl.FIPS() {
			t.Fatalf("FIPSProvider mismatch: want %v, got %v", openssl.FIPS(), fipsProv)
		}
	case openssl.MajorVersion() == 3:
		if (openssl.FIPS() || openssl.SymCryptProviderAvailable()) && !fipsProv {
			t.Fatal("FIPSProvider should be true when FIPS or SymCrypt is enabled")
		}
	}
}
```
Since both required fipsProv to be true, This covers both scenarios where FIPS is enabled or SymCryptProviderAvailable is available. Also more informative error message. WDYT? (maybe you can choose if else over switch in this case depending on your style of programming)

mertakman: I thought adding switch case is good for declaring second case only triggers when openssl version is 3(kinda serving as a documentation). But if its safe to assume anything non major version 1 will be automatically get as 3 by readers with little or no context, it is an argument too.

qmuntal: I see how the current implementation is somehow confusing. I've refactored the test to not depend on the OpenSSL version.

openssl.go
qmuntal: Good catch. The correct value is `fips=yes`, so I'll use it in the function comment.

mertakman (approved):

qmuntal:
I'm retiring this PR for now. It was primary targeted to workaround an issue when using AZL3 + SymCrypt +FIPS, and it may be finally fixed in AZL3.

PR #212, opened 2024-11-04
[ms-go1.23-support] Error out when encoding hash objects not implemented by the built-in providers

qmuntal:
This is a partial backport of #161 to the `ms-go1.23-support` branch.

We shouldn't read arbitrary memory and return it to the caller unless we know what we are decoding. This PR makes calls to `MarshalBinary` and `UnmarshalBinary` on hash objects to return an error when not using the built-in providers instead of attempting to marshal and unmarshal arbitrary memory, which normally end up in a panic while decoding the object.

Note that in #161 we went a little further and made the `NewSHAX` functions return an object that doesn't implement the encoding interfaces when not using the built-in providers. That change is too big to be backported.

### DIFF
--- a/hash_test.go
+++ b/hash_test.go
@@ -39,6 +39,27 @@ func cryptoToHash(h crypto.Hash) func() hash.Hash {
 	return nil
 }
 
+func TestHashNotMarshalable(t *testing.T) {
+	h := openssl.NewSHA256()
+	state, err := h.(encoding.BinaryMarshaler).MarshalBinary()
+	if err != nil {
+		t.Fatal(err)
+	}
+	*openssl.TestNotMarshalable = true
+	defer func() {
+		*openssl.TestNotMarshalable = false
+	}()
+
+	_, err = h.(encoding.BinaryMarshaler).MarshalBinary()
+	if err == nil {
+		t.Error("expected error")
+	}
+	err = h.(encoding.BinaryUnmarshaler).UnmarshalBinary(state)
+	if err == nil {
+		t.Error("expected error")
+	}
+}
+
 func TestHash(t *testing.T) {
 	msg := []byte("testing")
 	var tests = []struct {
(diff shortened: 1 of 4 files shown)


qmuntal:
macOS runners are failing because just today there is a scheduled burnout. See https://github.com/golang-fips/openssl/pull/213.

hash_test.go
```diff
@@ -39,6 +39,27 @@ func cryptoToHash(h crypto.Hash) func() hash.Hash {
...
+	state, err := h.(encoding.BinaryMarshaler).MarshalBinary()
+	if err != nil {
+		t.Fatal(err)
+	}
+	*openssl.TestNotMarshalable = true
+	defer func() {
+		*openssl.TestNotMarshalable = false
+	}()
+
+	_, err = h.(encoding.BinaryMarshaler).MarshalBinary()
```
dagood, line 55: Should a test like this be added to `v2` too?

hash_test.go
```diff
@@ -39,6 +39,27 @@ func cryptoToHash(h crypto.Hash) func() hash.Hash {
 	return nil
 }
 
+func TestHashNotMarshalable(t *testing.T) {
+	h := openssl.NewSHA256()
+	state, err := h.(encoding.BinaryMarshaler).MarshalBinary()
+	if err != nil {
+		t.Fatal(err)
```
dagood, line 48: Maybe it's worth mentioning in the error or a comment that this test assumes that SHA256 is always marshallable due to what we support in 1.23? (Or rather, what we don't support. 🙂)

dagood (approved):

hash_test.go
qmuntal: Good point, done!

hash_test.go
qmuntal: Probably, will do.

PR #213, opened 2024-11-04
Upgrade CI runner from macos-12 to macos-13

qmuntal:
MacOS CI runners are deprecated and will be removed soon: https://github.com/actions/runner-images/issues/10721.

In fact, macOS already entered the burnout period.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,7 +69,7 @@ jobs:
       matrix:
         go-version: [1.22.x, 1.23.x]
         openssl-version: [libcrypto.3.dylib]
-    runs-on: macos-14
+    runs-on: macos-13
     steps:
     - name: Install Go
       uses: actions/setup-go@v5
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -68,7 +68,7 @@ jobs:
       fail-fast: false
       matrix:
         go-version: [1.22.x, 1.23.x]
-        openssl-version: [libcrypto.3.dylib]
+        openssl-version: [libcrypto.dylib]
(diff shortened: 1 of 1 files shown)


gdams (approved):

qmuntal:
Grr, `libcrypto.3.dylib` does not exist in macos-14 (nor 13, nor 15...). In fact, the GH runner documentation says that those newer macOS images only ship with OpenSSL 1.1.1. OpenSSL 3 was probably included in macos-12 by mistake.

The problem is that there doesn't seem to be a `libcrypto1.1.dylib` file neither...

Maybe because just today macOS is migrating to OpenSSL 3 (https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md) and things a bit messy. Will wait till tomorrow.

dagood:
According to https://github.com/actions/runner-images/issues/10817 macOS 13 will get OpenSSL 3, so I suppose we should end up pointing at that (rather than 14) to get oldest possible coverage?

gdams:
Can we not `brew install openssl@3`?

gdams:
> According to [actions/runner-images#10817](https://github.com/actions/runner-images/issues/10817) macOS 13 will get OpenSSL 3, so I suppose we should end up pointing at that (rather than 14) to get oldest possible coverage?

actually yeah we may as well wait. Apparently "Image deployment will start on November 4 and will take 2-3 days."

qmuntal:
Finally found the right path to `libcrypto.3.dylib` 😅.

qmuntal:
@dagood PTAL

dagood (approved):

PR #214, opened 2024-11-04
[ms-go1.22-support] Error out when encoding hash objects not implemented by the built-in providers

qmuntal:
This is a backport of #212 to the `ms-go1.22-support` branch.

### DIFF
--- a/export_test.go
+++ b/export_test.go
@@ -1,3 +1,5 @@
 package openssl
 
 var ErrOpen = errOpen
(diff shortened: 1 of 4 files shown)


qmuntal:
macOS runners are failing because just today there is a scheduled burnout. See https://github.com/golang-fips/openssl/pull/213.

dagood (approved):

PR #219, opened 2024-11-13
Fix and improve SetFIPS

qmuntal:
#206 introduced a regression where `openssl.SetFIPS(false)` would always fail if there wasn't any FIPS-capable provider.

This PR fix that issue, adds some tests for `SetFIPS`, which was previously untested.

A benefit of the new implementation is that `OSSL_PROVIDER_try_load` is only called if the already loaded providers can't support the desired FIPS mode. It was previously unconditionally called, which was potentially unnecessary and could have undesired side-effects. For example, if the SymCrypt provider was already loaded and the built-in FIPS provider was not loaded but was installed in the system, then it would be loaded when calling `SetFIPS(true)`, ending with two conflicting FIPS providers available.

### DIFF
--- a/openssl.go
+++ b/openssl.go
@@ -103,6 +103,7 @@ var (
 	providerNameFips    = C.CString("fips")
 	providerNameDefault = C.CString("default")
 	propFIPS            = C.CString("fips=yes")
+	propNoFIPS          = C.CString("-fips")
 
 	algorithmSHA256 = C.CString("SHA2-256")
 )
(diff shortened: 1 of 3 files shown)


gdams (approved):

mertakman (approved):

openssl.go
```diff
@@ -103,6 +103,7 @@ var (
 	providerNameFips    = C.CString("fips")
 	providerNameDefault = C.CString("default")
 	propFIPS            = C.CString("fips=yes")
+	propNoFIPS          = C.CString("-fips")
```
dagood, line 106: I'm not able to find any docs on `-fips`, and the lack of symmetry vs. `fips=yes` is a bit strange. Where does this come from?

dagood: Not sure why it didn't show up in my initial searches, but found the meaning here: https://docs.openssl.org/3.0/man7/property/

dagood: Arguably the name could be `propIgnoreGlobalFIPS` to match the docs, but I don't know if that's meaningfully different.

derekparker (commented):
LGTM

PR #224, opened 2024-11-20
Implement ExpandHKDFOneShot

qmuntal:
Go 1.24 will most probably move the `golang.org/x/crypto/hkdf` package into the standard library: https://github.com/golang/go/issues/61477. The new API will be a little bit simpler, basically `hkdf.Expand` won't return a `io.Reader` but a slice of bytes as bis as the `keyLength` parameter.

Our `ExpandHKDF` function is not well fitted to implement the new API, as it also returns an `io.Reader` and internally allocated a buffer that might not have the same size as `keyLength`, which will end up causing unnecessary allocations.

This PR implements the new `ExpandHKDFOneShot` function that will be used to implement the new API.

### DIFF
--- a/hkdf.go
+++ b/hkdf.go
@@ -151,7 +151,7 @@ func ExtractHKDF(h func() hash.Hash, secret, salt []byte) ([]byte, error) {
 	}
 }
 
-// ExpandHKDFOneShot derives a key from the given hash, key, and optional context info,
+// ExpandHKDFOneShot derives a key from the given hash, key, and optional context info.
 func ExpandHKDFOneShot(h func() hash.Hash, pseudorandomKey, info []byte, keyLength int) ([]byte, error) {
 	if !SupportsHKDF() {
 		return nil, errUnsupportedVersion()
--- a/hkdf.go
+++ b/hkdf.go
@@ -151,6 +151,43 @@ func ExtractHKDF(h func() hash.Hash, secret, salt []byte) ([]byte, error) {
 	}
 }
 
+// ExpandHKDFOneShot derives a key from the given hash, key, and optional context info,
+func ExpandHKDFOneShot(h func() hash.Hash, pseudorandomKey, info []byte, keyLength int) ([]byte, error) {
+	if !SupportsHKDF() {
+		return nil, errUnsupportedVersion()
+	}
+
+	md, err := hashFuncToMD(h)
+	if err != nil {
+		return nil, err
+	}
+
+	out := make([]byte, keyLength)
+	switch vMajor {
+	case 1:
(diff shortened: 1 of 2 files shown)


hkdf.go
```diff
@@ -151,6 +151,43 @@ func ExtractHKDF(h func() hash.Hash, secret, salt []byte) ([]byte, error) {
...
+		ctx, err := newHKDFCtx3(md, C.GO_EVP_KDF_HKDF_MODE_EXPAND_ONLY, nil, nil, pseudorandomKey, info)
+		if err != nil {
+			return nil, err
+		}
+		defer C.go_openssl_EVP_KDF_CTX_free(ctx)
+		if C.go_openssl_EVP_KDF_derive(ctx, base(out), C.size_t(keyLength), nil) != 1 {
+			return nil, newOpenSSLError("EVP_KDF_derive")
+		}
+	default:
+		panic(errUnsupportedVersion())
```
dagood, line 186: Why `panic` vs. returning the error?

qmuntal: This is common practice in this repo. The supported version switches should be exhaustive so we know what to update when we start supporting a new version.

dagood: What difference does `return nil, errUnsupportedVersion()` vs. `panic(errUnsupportedVersion())` make to the switch's exhaustiveness?

FWIW, most uses of `errUnsupportedVersion` do use `panic`. Only tls1prf.go and hkdf.go return it, so maybe my confusion is in what situations it makes sense to `return` it rather than `panic`. (Maybe a doc comment on `errUnsupportedVersion` would make sense if it's worth explaining it for the future.)

<details><summary>(examples)</summary>

https://github.com/golang-fips/openssl/blob/4f0359e4649dbb346f83b1706519c8a01fc43b59/tls1prf.go#L50-L57

https://github.com/golang-fips/openssl/blob/4f0359e4649dbb346f83b1706519c8a01fc43b59/tls1prf.go#L15-L25

https://github.com/golang-fips/openssl/blob/a47a02278d7e3a14042576f160450e49072eaafc/hkdf.go#L17-L27

https://github.com/golang-fips/openssl/blob/a47a02278d7e3a14042576f160450e49072eaafc/hkdf.go#L113-L115

</details>

That said, since this isn't new, going ahead with approval. 😄

dagood (approved):

ueno (approved):

hkdf.go
qmuntal: > What difference does return nil, errUnsupportedVersion() vs. panic(errUnsupportedVersion()) make to the switch's exhaustiveness?

Go doesn't have exhaustive, so it is idiomatic to panic in the `default` case to at least get a runtime panic.

 Good catch about the inconsistency, will fix in a follow up PR.

PR #225, opened 2024-11-25
Update RSA invalid salt length error message

qmuntal:
Upstream just reworded the error message displayed when an invalid salt length is provided for RSA PSS ([source](https://github.com/golang/go/commit/93fcd8fb1882b55b3456aa753d32a2cf3d369b1c#diff-5add5ff057b0d06bb595f45e3e074652af0565974e8b2bb8022b27583295ae81L255)). That error message is used in an upstream test, so we need to follow suite.


ueno (approved):

dagood (approved):

qmuntal:
For https://github.com/microsoft/go/issues/1416.

PR #226, opened 2024-11-25
Validate ECDH keys

qmuntal:
Go 1.24 removed some ECDH key checks from the `crypto/ecdh` package and moved them to `crypto/internal/fips/ecdh`: https://github.com/golang/go/commit/566cf1c1083b26d4a15b94213c21142ecd9a8ca9#diff-e9f4458b20e24935551d73dc6b34e4007d91ca35e1e2c5b4d1e9f2a738d04e9cL79.

This means that the checks are no longer run in the the crypto backend layer, so we must instruct OpenSSL to perform that checks.

https://github.com/golang/go/commit/566cf1c1083b26d4a15b94213c21142ecd9a8ca9 also refactored the ECDH implementation so that `ecdh.PrivateKey.Public` is no longer lazy loaded, but generated when `ecdh.NewPrivateKey` is called. Following suite here, as it simplifies our implementation and makes key validation easier.

### DIFF
--- a/ecdh.go
+++ b/ecdh.go
@@ -165,7 +165,12 @@ func newECDHPkey1(nid C.int, bytes []byte, isPrivate bool) (pkey C.GO_EVP_PKEY_P
 		}
 	}
 	if C.go_openssl_EC_KEY_check_key(key) != 1 {
-		return nil, newOpenSSLError("EC_KEY_check_key")
+		// Match upstream error message.
+		if isPrivate {
+			return nil, errors.New("crypto/ecdh: invalid private key")
+		} else {
+			return nil, errors.New("crypto/ecdh: invalid public key")
+		}
 	}
 	return newEVPPKEY(key)
 }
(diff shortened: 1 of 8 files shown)


ecdh.go
```diff
@@ -20,49 +21,44 @@ func (k *PublicKeyECDH) finalize() {
...
-		return nil, errors.New("NewPublicKeyECDH: missing key")
+	if len(bytes) != 1+2*curveSize(curve) {
+		return nil, errors.New("NewPublicKeyECDH: wrong key length")
 	}
 	pkey, err := newECDHPkey(curve, bytes, false)
 	if err != nil {
 		return nil, err
 	}
-	k := &PublicKeyECDH{pkey, append([]byte(nil), bytes...)}
+	k := &PublicKeyECDH{pkey, slices.Clone(bytes)}
```
mertakman, line 40: good!

mertakman (approved):

evp.go
```diff
@@ -513,7 +510,34 @@ func newEvpFromParams(id C.int, selection C.int, params C.GO_OSSL_PARAM_PTR) (C.
 	}
 	var pkey C.GO_EVP_PKEY_PTR
 	if C.go_openssl_EVP_PKEY_fromdata(ctx, &pkey, selection, params) != 1 {
+		if vMinor < 2 {
```
dagood: Include major? Admittedly, mainly for clarity--seems unlikely to be a significant stumbling point if OpenSSL 4 comes along. 😄 

Edit: Oh, and the `<` doesn't match the comment that says 3.0.2 has the bug.

```suggestion
		if vMajor == 3 && vMinor <= 2 {
```

You have to explore a fair amount of context or know about this API in particular to know that the func will only be called when OpenSSL 3 is being used. Other funcs in this file do get called when using OpenSSL 1, so it seems reasonable to clarify in some way. `checkMajorVersion` would be another way, but has its cost.

dagood (approved):

gdams (approved):

dagood (approved):

qmuntal:
@derekparker the fix in this PR is blocking our upstream sync process. Could you take a look? Thanks.

qmuntal:
For https://github.com/microsoft/go/issues/1416.

PR #227, opened 2024-11-26
[ms-go1.23-support] Support MD5SHA1 in PKCS1 v1.5 signatures

qmuntal:
SymCrypt added support for using MD5SHA1 in PKCS1 v1.5 signatures, but we don't handle it correctly. This PR special case MD5SHA1 so it is used even when `EVP_MD_fetch` says it is not supported.

### DIFF
--- a/rsa_test.go
+++ b/rsa_test.go
@@ -158,7 +158,7 @@ func TestSignVerifyPKCS1v15_MD5SHA1(t *testing.T) {
 	if err != nil {
 		if openssl.FIPS() {
 			// This test is not supported in FIPS mode, but at least we
-			// can check that we don't panic (which we did before when using SOOSSL).
+			// can check that we don't panic (which we have before when using SCOSSL).
 			t.Skip("skipping test in FIPS mode")
 		}
 		t.Fatal(err)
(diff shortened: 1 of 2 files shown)


gdams (approved):

rsa_test.go
```diff
@@ -147,6 +147,28 @@ func TestSignVerifyPKCS1v15(t *testing.T) {
...
+
+	// Construct a fake MD5+SHA1 hash.
+	hashed := make([]byte, crypto.MD5.Size()+crypto.SHA1.Size())
+	hashed[0] = 0x30
+
+	signed, err := openssl.SignRSAPKCS1v15(priv, crypto.MD5SHA1, hashed)
+	if err != nil {
+		if openssl.FIPS() {
+			// This test is not supported in FIPS mode, but at least we
+			// can check that we don't panic (which we did before when using SOOSSL).
```
dagood:
```suggestion
			// can check that we don't panic (which we have before when using SCOSSL).
```

dagood (approved):

qmuntal:
Not sure if this worth backporting, as triggering this issue using the Go 1.23 `crypto/tls` package is not currently possible. It can only be triggered in synthetic tests. Will have to think more about it.

PR #229, opened 2024-11-29
Generate PKEYs using EVP_PKEY_Q_keygen

qmuntal:
OpenSSL supports one-shot PKEY generation using [EVP_PKEY_Q_keygen](https://docs.openssl.org/master/man3/EVP_PKEY_Q_keygen/).

Using this function from Go is a bit complicated because it has a variadic parameter, which is not directly supported by cgo (see https://github.com/golang/go/issues/975). I've instead defined the new `DEFINEFUNC_VARIADIC_3_0` macro that allows wrapping the same OpenSSL function multiple times with different numbers of arguments.

The code is now a bit more complex, but we shaved some cgo calls and also removed some [EVP_PKEY_CTX_ctrl](https://docs.openssl.org/master/man3/EVP_PKEY_CTX_ctrl) calls from this code path. This last bit is important because `EVP_PKEY_CTX_ctrl` is not recommended to be directly used in OpenSSL 3 (see quote below from docs), and I'm on a small quest to remove all direct calls to it from our codebase.

> Applications will not normally call EVP_PKEY_CTX_ctrl() directly but will instead call one of the algorithm specific functions below.

### DIFF
--- a/cmd/checkheader/main.go
+++ b/cmd/checkheader/main.go
@@ -227,6 +227,10 @@ func tryConvertDefineFunc(w io.Writer, l string, i int) bool {
 	if !strings.HasPrefix(l, "DEFINEFUNC") {
 		return false
 	}
+	if strings.HasPrefix(l, "DEFINEFUNC_VARIADIC") {
+		// Variadic functions are not supported.
+		return false
+	}
 	i1 := strings.IndexByte(l, '(')
 	// The first ")," match is always the end of the argument list parameter.
 	// We are not interested in the last parameter and parsing them would complicate the algorithm.
--- a/cmd/checkheader/main.go
+++ b/cmd/checkheader/main.go
@@ -228,7 +228,8 @@ func tryConvertDefineFunc(w io.Writer, l string, i int) bool {
 		return false
 	}
 	if strings.HasPrefix(l, "DEFINEFUNC_VARIADIC") {
-		// Variadic functions are not supported.
+		// Variadic functions are not supported. There is not enough
+		// information in the macro to create use it in writeDefineFunc.
 		return false
 	}
 	i1 := strings.IndexByte(l, '(')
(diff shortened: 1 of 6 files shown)


cmd/checkheader/main.go
```diff
@@ -227,6 +227,10 @@ func tryConvertDefineFunc(w io.Writer, l string, i int) bool {
 	if !strings.HasPrefix(l, "DEFINEFUNC") {
 		return false
 	}
+	if strings.HasPrefix(l, "DEFINEFUNC_VARIADIC") {
+		// Variadic functions are not supported.
```
dagood: Is this a point in time thing or is there something that makes it infeasible? Maybe it would be useful to mention why and/or link an issue.

shims.h
```diff
@@ -157,6 +157,12 @@ typedef void* GO_SHA_CTX_PTR;
 // DEFINEFUNC_RENAMED_3_0 acts like DEFINEFUNC but tries to load the function using the new name when using >= 3.x
 // and the old name when using 1.x. In both cases the function will have the new name.
 //
+// DEFINEFUNC_VARIADIC_3_0 defines a function that wraps an OpenSSL function with a different name and signature.
+// It should only be used for functions that can't be directly called from Go because their signature is not
+// compatible with cgo. The only known case are functions that take a variable number of arguments. See
+// https://github.com/golang/go/issues/975.
+// The process is aborted if the function can't be loaded when using 3.0.0 or higher.
```
dagood: I think this could be clearer...

* `VARIADIC` is a bit buried and that makes it harder to understand. I think it's worth focusing on why it's useful for that situation.
* The very first usage of this macro passes `EVP_PKEY_Q_keygen, EVP_PKEY_Q_keygen`, so it's odd to say that the name *is* different.
* Missing a bit of preamble.

```suggestion
// DEFINEFUNC_VARIADIC_3_0 acts like DEFINEFUNC but creates an alias with a more specific signature.
// This is necessary to call variadic functions (functions that accept a variable number of arguments)
// because variadic functions are not directly compatible with cgo. By defining a cgo-compatible alias
// for each desired signature, the C compiler handles the variadic arguments rather than cgo.
// Variadic functions are the only known incompatibility of this kind.
// If you use this macro for a different reason, consider renaming it to something more general first.
// See https://github.com/golang/go/issues/975.
// The process is aborted if the function can't be loaded when using 3.0.0 or higher.
```

dagood (approved):
Functionally, LGTM.

mertakman (approved):

PR #235, opened 2024-12-17
Support more hash types in RSA PKCS1v15 and PSS

qmuntal:
This PR expands the supported hash types, as reported by `openssl.SupportsHash`,  to increase the compatibility of the RSA PCKS1v15 and PSS sign and verify functions with the standard library. It is now possible to use those algorithms with `crypto.RIPEMD160`, `crypto.SHA512_224` and `crypto.SHA512_256`, if the OpenSSL provider (or legacy engine) supports it.

While here, clean up `cryptoHashToMD` and `SupportsHash`. The former doesn't need to check if the given hash is supported or not.

### DIFF
--- a/rsa_test.go
+++ b/rsa_test.go
@@ -194,26 +194,26 @@ func TestRSAEncryptDecryptOAEP_WrongLabel(t *testing.T) {
 	}
 }
 
+// These are all the hashes supported by Go's crypto/rsa package
+// as off Go 1.24.
+var stdHashes = [...]crypto.Hash{
+	crypto.MD5SHA1,
+	crypto.MD5,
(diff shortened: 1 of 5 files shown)


rsa_test.go
```diff
@@ -193,7 +194,61 @@ func TestRSAEncryptDecryptOAEP_WrongLabel(t *testing.T) {
 	}
 }
 
+// These are all the hashes supported by Go's crypto/rsa package
+// as off Go 1.24.
```
dagood:
```suggestion
// as of Go 1.24.
```

dagood (approved):

gdams (approved):

PR #238, opened 2024-12-20
Speed up NewShaX

qmuntal:
In many places we create a hash object, e.g. by calling `openssl.NewSHA256()`, just to get the type of hash and map it to an OpenSSL hash identifier. Then the original hash is discarded. These pattern is common enough to worth optimizing.

This PR refactors the `evpHash` implementation so that the hash `EVP_MD_CTX` context is created and initialized lazily. These are the speed gains:

```
goos: windows
goarch: amd64
             │    old.txt     │               new.txt               │
             │     sec/op     │   sec/op     vs base                │
NewSHA256-12   1533.50n ± 27%   50.39n ± 8%  -96.71% (p=0.000 n=10)

             │  old.txt   │            new.txt             │
             │    B/op    │    B/op     vs base            │
NewSHA256-12   24.00 ± 0%   24.00 ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal

             │  old.txt   │            new.txt             │
             │ allocs/op  │ allocs/op   vs base            │
NewSHA256-12   1.000 ± 0%   1.000 ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal
```

### DIFF
--- a/hash.go
+++ b/hash.go
@@ -227,27 +227,42 @@ func newEvpHash(ch crypto.Hash) *evpHash {
 	if alg == nil {
 		panic("openssl: unsupported hash function: " + strconv.Itoa(int(ch)))
 	}
-	ctx := C.go_openssl_EVP_MD_CTX_new()
-	if C.go_openssl_EVP_DigestInit_ex(ctx, alg.md, nil) != 1 {
-		C.go_openssl_EVP_MD_CTX_free(ctx)
-		panic(newOpenSSLError("EVP_DigestInit_ex"))
-	}
-	ctx2 := C.go_openssl_EVP_MD_CTX_new()
-	h := &evpHash{
-		alg:  alg,
-		ctx:  ctx,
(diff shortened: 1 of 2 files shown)


gdams (approved):

hash.go
```diff
@@ -360,6 +384,8 @@ func (d *evpHash) MarshalBinary() ([]byte, error) {
 }
 
 func (d *evpHash) AppendBinary(buf []byte) ([]byte, error) {
+	defer runtime.KeepAlive(d)
```
dagood, line 387: Why is this KeepAlive necessary? (And why now?)

dagood (approved):

hash.go
qmuntal: Not really necessary, as `d` is referenced after accessing the context. But I get nervous when I see a missing KeepAlive, as a refactor could break the code.

PR #239, opened 2024-12-20
Remove MD5SHA1 OpenSSL 1.0.2 backport

qmuntal:
We only implemented `crypto.MD5SHA1` to support TLS 1.0/1.1 handshakes, which uses that hash in RSA PKCS1 v15 operations. Unfortunately, OpenSSL 1.0.2 doesn't provide `EVP_md5_sha1`, so we had to manually port it thinking that we could make that work by doing that.

The reality is that our backport doesn't work, as demonstrated by the tests added in https://github.com/golang-fips/openssl/pull/235. Given it complexity, and that OpenSSL 1.0.2 is long past it's EOL date, we better remove it.


gdams (approved):

ueno (approved):
+1 for the removal.

PR #245, opened 2025-01-10
Implement FIPSCapable

qmuntal:
When using OpenSSL 3, `openssl.FIPS` prove that FIPS mode is enabled by checking that the default properties contain `fips=yes` and that SHA-256 can be fetched with the default properties. This behavior is useful to know if there is an explicit intent to only run in FIPS mode.

On the other hand, `openssl.FIPS` doesn't cover the use case of wanting to know whether the provider configured by default is FIPS-capable regardless if the default properties. In fact, most times users don't care about the default properties, they just want to ensure they are running in FIPS mode, aka the default provider is FIPS capable. 

This PR adds the `openssl.FIPSCapable` function to cover this new use case.

For https://github.com/microsoft/go/issues/1445.

### DIFF
--- a/openssl.go
+++ b/openssl.go
@@ -134,6 +134,28 @@ func FIPS() bool {
 	}
 }
 
+// FIPSCapable returns true if the provider used by the default matches the `fips=yes` query.
+// Note that this function can return true even if [FIPS] returns false, because [FIPS] also
+// checks whether the default properties contain `fips=yes`.
+// It will always return true for OpenSSL 3 if [FIPS] returns true.
+// When using OpenSSL 1, this function always returns the same value as [FIPS].
+func FIPSCapable() bool {
+	if FIPS() {
+		return true
+	}
+	if vMajor == 3 {
+		// Load the provider with and without the `fips=yes` query.
+		// If the providers are the same, then the default provider is FIPS-capable.
+		provFIPS := sha256Provider(propFIPS)
+		if provFIPS == nil {
+			return false
+		}
+		provDefault := sha256Provider(nil)
+		return provFIPS == provDefault
+	}
+	return false
+}
+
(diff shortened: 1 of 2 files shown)


openssl.go
```diff
@@ -134,6 +134,28 @@ func FIPS() bool {
...
+// checks whether the default properties contain `fips=yes`.
+// It will always return true for OpenSSL 3 if [FIPS] returns true.
+// When using OpenSSL 1, this function always returns the same value as [FIPS].
+func FIPSCapable() bool {
+	if FIPS() {
+		return true
+	}
+	if vMajor == 3 {
+		// Load the provider with and without the `fips=yes` query.
+		// If the providers are the same, then the default provider is FIPS-capable.
```
dagood, line 157: I'm not sure this matches what I'd expect based on the word "capable". A system that has a `fips=yes` provider but doesn't use that provider by default is still "capable", isn't it? (Perhaps it's configured to use a faster provider when `fips=yes` isn't required.) I'm not totally sure what this API would actually be used for--maybe documenting a use case in the func doc would tie it all together.

Checking the provider for sha256 in particular also could make it a bit complicated, right? If the default properties don't contain `fips=yes`, IIRC some other algorithm might fall back to a non-FIPS provider. (I know this is a standard assumption elsewhere, but it seems worth considering in this particular context depending on how the function will be used.)

I know we're not aiming for a 100% accurate general-purpose detection, though. My mind is probably wandering outside the configurations that are known to exist on our target systems. 😄

qmuntal: I'll improve the func docs a bit more 👍 .

dagood: Thanks, I like the example and note about heuristics in particular.

dagood (approved):

gdams (approved):

PR #246, opened 2025-01-14
Add wrapper for `EVP_BPE_scrypt`

harrybrwn:
Pretty simple, just a wrapper function around `EVP_BPE_scrypt` with some simple tests.

### DIFF
--- a/scrypt.go
+++ b/scrypt.go
@@ -1,0 +1,37 @@
+//go:build !cmd_go_bootstrap
+
+package openssl
+
+// #include "goopenssl.h"
+import "C"
+import (
+	"unsafe"
+)
+
+func Scrypt(password string, salt []byte, N, r, p, maxmem, keylen uint64) ([]byte, error) {
+	cpassword := C.CString(password)
+	defer C.free((unsafe.Pointer)(cpassword))
+
+	csalt := C.CBytes(salt)
+	defer C.free((unsafe.Pointer)(csalt))
+
+	key := C.malloc(C.size_t(keylen))
+	defer C.free(key)
+
+	res := C.go_openssl_EVP_PBE_scrypt(
+		cpassword,
+		C.size_t(len(password)),
+		(*C.uchar)(csalt),
+		C.size_t(len(salt)),
+		C.uint64_t(N),
+		C.uint64_t(r),
+		C.uint64_t(p),
+		C.uint64_t(maxmem),
+		(*C.uchar)(key),
+		C.size_t(keylen),
+	)
+	if res != 1 {
+		return nil, newOpenSSLError("EVP_PBE_scrypt")
+	}
+	return C.GoBytes(key, C.int(keylen)), nil
+}
--- a/scrypt.go
+++ b/scrypt.go
@@ -8,20 +8,19 @@ import (
 	"unsafe"
 )
 
-func Scrypt(password string, salt []byte, N, r, p, maxmem, keylen uint64) ([]byte, error) {
+func Scrypt(password string, salt []byte, N, r, p, maxMem, keyLen uint64) ([]byte, error) {
 	cpassword := C.CString(password)
 	defer C.free((unsafe.Pointer)(cpassword))
 
 	csalt := C.CBytes(salt)
 	defer C.free((unsafe.Pointer)(csalt))
 
-	key := C.malloc(C.size_t(keylen))
-	defer C.free(key)
+	key := make([]byte, keylen)
 
 	res := C.go_openssl_EVP_PBE_scrypt(
-		cpassword,
+		base(unsafe.StringData(password)),
 		C.size_t(len(password)),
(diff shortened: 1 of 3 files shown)


scrypt.go
```diff
@@ -0,0 +1,37 @@
...
+	defer C.free((unsafe.Pointer)(cpassword))
+
+	csalt := C.CBytes(salt)
+	defer C.free((unsafe.Pointer)(csalt))
+
+	key := C.malloc(C.size_t(keylen))
+	defer C.free(key)
+
+	res := C.go_openssl_EVP_PBE_scrypt(
+		cpassword,
```
qmuntal: The password content is not being modified, so there is no need to create a copy of it. `unsafe` is our friend here:

```suggestion
		base(unsafe.StringData(password)),
```

scrypt.go
```diff
@@ -0,0 +1,37 @@
...
+		C.size_t(len(password)),
+		(*C.uchar)(csalt),
```
qmuntal: Same as with `password`, no need to copy `salt` as it is not being modified:

```suggestion
		base(salt),
```

scrypt.go
```diff
@@ -0,0 +1,37 @@
...
+
+func Scrypt(password string, salt []byte, N, r, p, maxmem, keylen uint64) ([]byte, error) {
+	cpassword := C.CString(password)
+	defer C.free((unsafe.Pointer)(cpassword))
+
+	csalt := C.CBytes(salt)
+	defer C.free((unsafe.Pointer)(csalt))
+
+	key := C.malloc(C.size_t(keylen))
+	defer C.free(key)
```
qmuntal: `key` can directly be allocated in Go memory and passed to `go_openssl_EVP_PBE_scrypt` using `base(key)`.

```suggestion
	key := make([]byte, keylen)
```

shims.h
```diff
@@ -410,4 +410,4 @@ DEFINEFUNC_3_0(int, EVP_KDF_CTX_set_params, (GO_EVP_KDF_CTX_PTR ctx, const GO_OS
 DEFINEFUNC_3_0(void, EVP_KDF_CTX_free, (GO_EVP_KDF_CTX_PTR ctx), (ctx)) \
 DEFINEFUNC_3_0(size_t, EVP_KDF_CTX_get_kdf_size, (GO_EVP_KDF_CTX_PTR ctx), (ctx)) \
 DEFINEFUNC_3_0(int, EVP_KDF_derive, (GO_EVP_KDF_CTX_PTR ctx, unsigned char *key, size_t keylen, const GO_OSSL_PARAM_PTR params), (ctx, key, keylen, params)) \
-
+DEFINEFUNC_1_1_1(int, EVP_PBE_scrypt, (const char *pass, size_t passlen, const unsigned char *salt, size_t saltlen, uint64_t N, uint64_t r, uint64_t p, uint64_t maxmem, unsigned char *key, size_t keylen), (pass, passlen, salt, saltlen, N, r, p, maxmem, key, keylen))
```
qmuntal, line 413: `EVP_PBE_scrypt` is defined since OpenSSL 3.0

```suggestion
DEFINEFUNC_3_0(int, EVP_PBE_scrypt, (const char *pass, size_t passlen, const unsigned char *salt, size_t saltlen, uint64_t N, uint64_t r, uint64_t p, uint64_t maxmem, unsigned char *key, size_t keylen), (pass, passlen, salt, saltlen, N, r, p, maxmem, key, keylen)) \

```

scrypt.go
```diff
@@ -0,0 +1,37 @@
...
+
+package openssl
+
+// #include "goopenssl.h"
+import "C"
+import (
+	"unsafe"
+)
+
+func Scrypt(password string, salt []byte, N, r, p, maxmem, keylen uint64) ([]byte, error) {
```
qmuntal: Nit: use Go  naming conventions

```suggestion
func Scrypt(password string, salt []byte, N, r, p, maxMem, keyLen uint64) ([]byte, error) {
```

scrypt.go
```diff
@@ -0,0 +1,37 @@
+//go:build !cmd_go_bootstrap
+
+package openssl
+
+// #include "goopenssl.h"
+import "C"
+import (
+	"unsafe"
+)
+
```
qmuntal, line 10: `EVP_PBE_scrypt` is supported since OpenSSL 3, and maybe not for all providers. Please implement the `SupportsSscrypt() bool` function. The implementation will be almost identical to https://github.com/golang-fips/openssl/blob/eb155dada337b0de53873aa42e89724ad59aa768/pbkdf2.go#L15, replacing `PBKDF2` with `SCRYPT` and returning `false` in the OpenSSL 1 case.

Then use this function in the new test cases.

qmuntal (changes requested):
Thanks for submitting. Adding Scrypt support seems like a good addition. Added some comments.

PR #254, opened 2025-02-21
Implement GCM Seal and Open in Go

qmuntal:
The less C code we have, the better. This PR moves the GCM Seal/Open implementation from C to Go. It was previously written in C to avoid allocations, but that's no longer need thanks to the `noescape` and `nocallback` directive.

Note that I haven't ported some special-casing for OpenSSL 1.0.2, as it is no longer supported.

These benchmark run demonstrates that there is not performance regression:

```
goos: linux
goarch: amd64
              │    sec/op    │    sec/op     vs base               │
AESGCM_Open-4   1.708µ ± 17%   1.814µ ±  6%       ~ (p=0.190 n=10)
AESGCM_Seal-4   1.876µ ± 21%   1.771µ ± 15%       ~ (p=0.224 n=10)
geomean         1.790µ         1.792µ        +0.13%

              │    old.txt    │               new.txt                │
              │      B/s      │      B/s       vs base               │
AESGCM_Open-4   35.74Mi ± 14%   33.65Mi ±  5%       ~ (p=0.190 n=10)
AESGCM_Seal-4   32.57Mi ± 17%   34.47Mi ± 13%       ~ (p=0.224 n=10)
geomean         34.12Mi         34.06Mi        -0.18%

              │     B/op     │    B/op     vs base                 │
AESGCM_Open-4   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
AESGCM_Seal-4   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
geomean                    ²               +0.00%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean

              │  allocs/op   │ allocs/op   vs base                 │
AESGCM_Open-4   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
AESGCM_Seal-4   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
geomean                    ²               +0.00%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean
```

### DIFF
--- a/cipher.go
+++ b/cipher.go
@@ -381,7 +381,7 @@ func (g *cipherGCM) Overhead() int {
 	return gcmTagSize
 }
 
-func (g *cipherGCM) Seal(dst, nonce, plaintext, additionalData []byte) []byte {
+func (g *cipherGCM) Seal(dst, nonce, plaintext, aad []byte) []byte {
 	if len(nonce) != gcmStandardNonceSize {
 		panic("cipher: incorrect nonce length given to GCM")
 	}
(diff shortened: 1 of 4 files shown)


gdams (approved):

cipher.go
```diff
@@ -457,19 +457,30 @@ func (g *cipherGCM) Seal(dst, nonce, plaintext, additionalData []byte) []byte {
...
+		panic(newOpenSSLError("EVP_CIPHER_CTX_ctrl"))
 	}
 	runtime.KeepAlive(g)
 	return ret
 }
 
 var errOpen = errors.New("cipher: message authentication failed")
 
-func (g *cipherGCM) Open(dst, nonce, ciphertext, additionalData []byte) ([]byte, error) {
+func (g *cipherGCM) Open(dst, nonce, ciphertext, aad []byte) (_ []byte, err error) {
```
mertakman, line 483: nitpick:why not just used "add" for additional data instead of "aad"? other than that,its okey.

mertakman (approved):

cipher.go
qmuntal: `aad` is the common idiom for this variable. It stands for `Additional Authenticated Data`.

PR #256, opened 2025-02-24
Define int constants in Go instead of in C

qmuntal:
Defining OpenSSL constants in Go rather than C makes the codebase a bit easier to maintain as there is less cgo involved.


gdams (approved):

cmd/checkheader/main.go
```diff
@@ -169,25 +169,22 @@ func tryConvertTypedef(w io.Writer, l string) bool {
...
-// the enum contained in the line l
+// tryConvertGoConst adds a static check which verifies that
+// the const contained in the line l
 // matches the corresponding OpenSSL value.
-// Only enum names starting with GO_ are converted.
-func tryConvertEnum(w io.Writer, l string) {
-	if !strings.HasPrefix(l, "GO_") {
+// Only const names starting with _ are converted.
+func tryConvertGoConst(w io.Writer, l string) {
+	if !strings.HasPrefix(l, "_") || strings.HasSuffix(l, "// nocheck") {
```
dagood: Maybe `//checkheader:nocheck` or `//checkheader:ignore` rather than `// nocheck` to [officially](https://tip.golang.org/doc/comment#syntax) make it a directive comment?

(Just for readability--AFAIK even if we used the Go parser/AST packages, they don't actually have a built-in way to look for these.)

cmd/checkheader/main.go
```diff
@@ -113,14 +113,14 @@ func generate(header string) (string, error) {
...
-				tryConvertEnum(&b, l)
+			if !strings.HasPrefix(l, ")") {
+				tryConvertGoConst(&b, l)
 			} else {
 				enum = false
 			}
 			continue
 		}
-		if strings.HasPrefix(l, "enum {") || strings.HasPrefix(l, "typedef enum {") {
+		if l == "const (" {
```
dagood: This `"const ( // no_check" == "const ("` check is actually what skips the const blocks, not the `strings.HasSuffix(l, "// nocheck")` check in `tryConvertGoConst`... seems confusing.

Maybe change this line to look for a `const (` prefix and lack of `nocheck` suffix, so the intent is clearer?

mertakman (approved):

cmd/checkheader/main.go
qmuntal: Like the idea. Will use `//checkheader:ignore`.

cmd/checkheader/main.go
qmuntal: Will do!

PR #257, opened 2025-02-24
Don't use encoding/binary

qmuntal:
Upstream Go doesn't like encoding/binary to be in a dependency of a crypto package, as it is quite heavy because it uses reflect. Remove it.

### DIFF
--- a/openssl.go
+++ b/openssl.go
@@ -404,7 +404,7 @@ func versionAtOrAbove(major, minor, patch uint) bool {
 	return vMajor > major || (vMajor == major && vMinor > minor) || (vMajor == major && vMinor == minor && vPatch >= patch)
 }
 
-func bigUint64(b []byte) uint64 {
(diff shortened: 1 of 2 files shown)


openssl.go
```diff
@@ -404,3 +403,9 @@ func CheckLeaks() {
 func versionAtOrAbove(major, minor, patch uint) bool {
 	return vMajor > major || (vMajor == major && vMinor > minor) || (vMajor == major && vMinor == minor && vPatch >= patch)
 }
+
+func bigUint64(b []byte) uint64 {
```
dagood: Seems too easy to mix up with BigInt:

```suggestion
func bigEndianUint64(b []byte) uint64 {
```

dagood (approved):
(Same name change as CNG side.)

gdams (approved):

PR #258, opened 2025-02-26
Remove hash hacks

qmuntal:
We can use a `runtime.Pinner` and the `noescape`/`nocallback` directives to replace some hacky logic that we implemented in the several hash functions to avoid allocations and to workaround a `cgoCheckPointer` false positive (tested in `TestCgo`).

Performance is not affected:

```
goos: linux
goarch: amd64
                │   old.txt    │              new.txt               │
                │    sec/op    │   sec/op     vs base               │
SHA256-4          814.0n ±  3%   792.2n ± 6%       ~ (p=0.105 n=10)
SHA256OneShot-4   585.8n ± 15%   607.1n ± 4%       ~ (p=0.089 n=10)
geomean           690.5n         693.5n       +0.42%

                │    old.txt    │               new.txt               │
                │      B/s      │     B/s       vs base               │
SHA256-4          9.370Mi ±  2%   9.632Mi ± 6%       ~ (p=0.093 n=10)
SHA256OneShot-4   13.02Mi ± 18%   12.56Mi ± 4%       ~ (p=0.089 n=10)
geomean           11.05Mi         11.00Mi       -0.41%

                │     B/op     │    B/op     vs base                 │
SHA256-4          64.00 ± 0%     64.00 ± 0%       ~ (p=1.000 n=10) ¹
SHA256OneShot-4   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
geomean                      ²               +0.00%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean

                │  allocs/op   │ allocs/op   vs base                 │
SHA256-4          1.000 ± 0%     1.000 ± 0%       ~ (p=1.000 n=10) ¹
SHA256OneShot-4   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
geomean                      ²               +0.00%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean
```

### DIFF
--- a/openssl.go
+++ b/openssl.go
@@ -232,31 +232,7 @@ func proveSHA256(props cString) bool {
 	return sha256Provider(props) != nil
 }
 
-// noescape hides a pointer from escape analysis. noescape is
-// the identity function but escape analysis doesn't think the
-// output depends on the input. noescape is inlined and currently
-// compiles down to zero instructions.
-// USE CAREFULLY!
-//
-//go:nosplit
-func noescape(p unsafe.Pointer) unsafe.Pointer {
-	x := uintptr(p)
-	return unsafe.Pointer(x ^ 0)
-}
-
 var zero byte
-
-// addr converts p to its base addr, including a noescape along the way.
-// If p is nil, addr returns a non-nil pointer, so that the result can always
-// be dereferenced.
-//
-//go:nosplit
-func addr(p []byte) *byte {
-	if len(p) == 0 {
-		return &zero
-	}
-	return (*byte)(noescape(unsafe.Pointer(&p[0])))
-}
 
 // baseNeverEmpty returns the address of the underlying array in b.
 // If b has zero length, it returns a pointer to a zero byte.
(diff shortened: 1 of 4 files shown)


gdams (approved):

openssl.go
```diff
@@ -267,20 +243,22 @@ func baseNeverEmpty(b []byte) *C.uchar {
 	return (*C.uchar)(unsafe.Pointer(&b[0]))
 }
 
-// base returns the address of the underlying array in b,
-// being careful not to panic when b has zero length.
-func base(b []byte) *C.uchar {
+// pbase returns the address of the underlying array in b.
```
dagood: Should this also have the carefulness comment? If it doesn't, it seems like by contrast `pbase` *won't* be careful while `base` would be careful.

```suggestion
// pbase returns the address of the underlying array in b,
// being careful not to panic when b has zero length.
```

While here, maybe add a comment to `sbase` too?

dagood (approved):

Issue #231, opened 2024-12-09
Error handling is not thead-safe


qmuntal:
We use [ERR_get_error](https://docs.openssl.org/3.3/man3/ERR_get_error/#synopsis) to retrieve the error that caused a given function call to fail. This function docs says this:

> ERR_get_error() returns the earliest error code from the thread's error queue and removes the entry. 

The problem is that cgo calls can be rescheduled on any thread when switching back from the system stack to the goroutine stack, The Go scheduler does this to avoid goroutines having to wait for the original thread to be available.

This means that our call to `ERR_get_error` might not return the correct error, or even no error at all.

There are two ways to solve this:

- Calling `runtime.LockOSThread` and `defer runtime.UnlockOSThread` before every cgo call so that the goroutine is always recheduled to the same thread.
- Replacing all OpenSSL cgo calls for a custom C wrapper function that calls the OpenSSL C function and also `ERR_get_error` if that fails. Note that this will work because a goroutine can't be rescheduled while in the system stack.

The first option is the easiest one, but will impact the performance of our OpenSSL backend even in the no-error cases. I would prefer to investigate how we could implement the second option, which wouldn't have any performance impact. Note that both options will require big refactors, as we will need to wrap existing cgo calls into either a C or a Go function (or even both).

PR #259, opened 2025-02-28
Autogenerate C and Go wrappers

qmuntal:
Until now we have been generating the C function headers. the C functions `dlsym` calls, and the C wrappers using good old C macros. C macros have the advantage of not needing to implement a parser for a new domain specific language (aka DSL) and to be pretty compact. On the other hand, they are difficult to troubleshoot if there is any issue with them, the error message is long and cryptic, and, most important, they have its limit on what can be sanely implemented with them.

This PR sets the ground for many improvements to come by dropping the C macro approach and instead use implement a small application (`mkcgo`) that can read C function definitions (limited to what is necessary to support OpenSSL) and then generates all the C stubs plus a handy Go wrappers over them.

This PR doesn't have any behavior change, it's only syntactic changes. It does add ~3000 line of code, but almost all of them are from autogenerated code.


mertakman (approved):

dagood:
For context, based on a conversation earlier, one of the major things this will make reasonable to implement is a good fix for:

* https://github.com/golang-fips/openssl/issues/231

cmd/mkcgo/main.go
```diff
@@ -0,0 +1,149 @@
...
+		for _, d := range [][]byte{data, cbuf.Bytes()} {
+			if _, err = os.Stdout.Write(d); err != nil {
+				log.Fatal(err)
+			}
+		}
+	} else {
+		err = os.WriteFile(*fileName, data, 0644)
+		if err == nil {
+			cfileName := strings.TrimSuffix(*fileName, ".go") + ".c"
+			err = os.WriteFile(cfileName, cbuf.Bytes(), 0644)
```
dagood:
```suggestion
		err = os.WriteFile(*fileName, data, 0o644)
		if err == nil {
			cfileName := strings.TrimSuffix(*fileName, ".go") + ".c"
			err = os.WriteFile(cfileName, cbuf.Bytes(), 0o644)
```

shims.h
```diff
@@ -1,82 +1,42 @@
...
-typedef void* GO_EVP_MAC_PTR;
-typedef void* GO_EVP_MAC_CTX_PTR;
-typedef void* GO_OSSL_PARAM_BLD_PTR;
-typedef void* GO_OSSL_PARAM_PTR;
-typedef void* GO_CRYPTO_THREADID_PTR;
-typedef void* GO_EVP_SIGNATURE_PTR;
-typedef void* GO_DSA_PTR;
-typedef void* GO_EVP_KDF_PTR;
-typedef void* GO_EVP_KDF_CTX_PTR;
+// Tags used to determine which OpenSSL version the function is available in:
```
dagood: Worth describing what this list is for, what `mkcgo` refers to?

shims.h
```diff
@@ -93,222 +53,224 @@ typedef void* GO_SHA_CTX_PTR;
...
-DEFINEFUNC_3_0(void, EVP_KDF_free, (GO_EVP_KDF_PTR kdf), (kdf)) \
-DEFINEFUNC_3_0(GO_EVP_KDF_CTX_PTR, EVP_KDF_CTX_new, (GO_EVP_KDF_PTR kdf), (kdf)) \
-DEFINEFUNC_3_0(int, EVP_KDF_CTX_set_params, (GO_EVP_KDF_CTX_PTR ctx, const GO_OSSL_PARAM_PTR params), (ctx, params)) \
-DEFINEFUNC_3_0(void, EVP_KDF_CTX_free, (GO_EVP_KDF_CTX_PTR ctx), (ctx)) \
-DEFINEFUNC_3_0(size_t, EVP_KDF_CTX_get_kdf_size, (GO_EVP_KDF_CTX_PTR ctx), (ctx)) \
-DEFINEFUNC_3_0(int, EVP_KDF_derive, (GO_EVP_KDF_CTX_PTR ctx, unsigned char *key, size_t keylen, const GO_OSSL_PARAM_PTR params), (ctx, key, keylen, params)) \
+/*[[mkcgo]]*/ void ERR_error_string_n(unsigned long e, char *buf, size_t len);
+/*[[mkcgo]]*/ void ERR_clear_error(void);
+/*[[mkcgo::tag("legacy_1")]]*/ unsigned long ERR_get_error_line(const char **file, int *line);
+/*[[mkcgo::tag("3")]]*/ unsigned long ERR_get_error_all(const char **file, int *line, const char **func, const char **data, int *flags);
```
dagood: Maybe it's something to get used to (or doesn't matter because of the "as needed" nature of making edits to this list), but it seems hard to read. A suffix might help: less variability in where the function name actually ends up and more familiar--allows `//` and positioned like Go struct tags. And maybe newlines help even more:

```c
// CRYPTO_malloc argument num changes from int to size_t in OpenSSL 1.1.0, 
// and CRYPTO_free has file and line arguments added. 
// Exclude them from headercheck tool when using previous OpenSSL versions. 
void *CRYPTO_malloc(size_t num, const char *file, int line); // [[mkcgo]] 
void CRYPTO_free(void *str, const char *file, int line); // [[mkcgo]] 

int OPENSSL_init_crypto(uint64_t ops, const _OPENSSL_INIT_SETTINGS_PTR settings); // [[mkcgo]] 

int FIPS_mode(void); // [[mkcgo::tag("legacy_1")]] 
int FIPS_mode_set(int r); // [[mkcgo::tag("legacy_1")]] 

int EVP_default_properties_is_fips_enabled(_OSSL_LIB_CTX_PTR libctx); // [[mkcgo::tag("3")]] 
int EVP_default_properties_enable_fips(_OSSL_LIB_CTX_PTR libctx, int enable); // [[mkcgo::tag("3")]] 

int OSSL_PROVIDER_available(_OSSL_LIB_CTX_PTR libctx, const char *name); // [[mkcgo::tag("3")]] 
_OSSL_PROVIDER_PTR OSSL_PROVIDER_try_load(_OSSL_LIB_CTX_PTR libctx, const char *name, int retain_fallbacks); // [[mkcgo::tag("3")]] 
const char *OSSL_PROVIDER_get0_name(const _OSSL_PROVIDER_PTR prov); // [[mkcgo::tag("3")]] 

_EVP_MD_PTR EVP_MD_fetch(_OSSL_LIB_CTX_PTR ctx, const char *algorithm, const char *properties); // [[mkcgo::tag("3")]] 
void EVP_MD_free(_EVP_MD_PTR md); // [[mkcgo::tag("3")]] 
const char *EVP_MD_get0_name(const _EVP_MD_PTR md); // [[mkcgo::tag("3")]] 
const _OSSL_PROVIDER_PTR EVP_MD_get0_provider(const _EVP_MD_PTR md); // [[mkcgo::tag("3")]] 
int EVP_MD_get_type(const _EVP_MD_PTR md); // [[mkcgo::tag("3")]] 
int EVP_MD_size(const _EVP_MD_PTR md); // [[mkcgo::tag("legacy_1")]] 
int EVP_MD_get_size(const _EVP_MD_PTR md); // [[mkcgo::tag("3")]] 
int EVP_MD_block_size(const _EVP_MD_PTR md); // [[mkcgo::tag("legacy_1")]] 
int EVP_MD_get_block_size(const _EVP_MD_PTR md); // [[mkcgo::tag("3")]] 
```

cmd/checkheader/main.go
```diff
@@ -197,108 +196,74 @@ func tryConvertGoConst(w io.Writer, l string) {
...
-	}
-	if !strings.HasPrefix(l, "DEFINEFUNC") {
+	if !strings.HasPrefix(l, "/*[[mkcgo") {
 		return false
 	}
-	if strings.HasPrefix(l, "DEFINEFUNC_VARIADIC") {
+	if strings.Contains(l, "[[mkcgo::variadic(") {
 		// Variadic functions are not supported. There is not enough
-		// information in the macro to create use it in writeDefineFunc.
+		// information to create use it in writeDefineFunc.
```
dagood:
```suggestion
		// information to create it in writeDefineFunc.
```

cmd/checkheader/main.go
```diff
@@ -197,108 +196,74 @@ func tryConvertGoConst(w io.Writer, l string) {
...
 			return
 		}
-		fnret, fnname, fnargs := args[0], args[1], args[2]
+		idx := strings.LastIndexByte(argsAndName, ' ')
+		if idx < 0 {
+			log.Printf("can't retrive return type in line: %s\n", l)
+			return
+		}
+		fnret := argsAndName[:idx]
+		fnname := argsAndName[idx+1:]
```
dagood: Would it make sense to share the parser with `mkcgo`? (Index-based logic in particular usually seems fiddly and panic-happy to me, enough to be worth sharing to at least reduce the amount of it.)

shims.h
```diff
@@ -93,222 +53,224 @@ typedef void* GO_SHA_CTX_PTR;
 // #if OPENSSL_VERSION_NUMBER < 0x10100000L
 // #include <openssl/bn.h>
 // #endif
```
dagood: Now that we have two processors looking at this file, worth making it more explicit which processor cares about these lines (and that one of them does care at all!) by using a prefix or wrapping it somehow?

cmd/checkheader/main.go
qmuntal: Good suggestion. Done 😄.

shims.h
qmuntal: Good point. I've added a comment to explain why this includes are here. It think this is enough.

shims.h
qmuntal: I've added some more comments 👍

shims.h
qmuntal: I've made the `[[mkcgo]]` implicit and switched to the good old GCC [__attribute__](https://gcc.gnu.org/onlinedocs/gcc/Function-Attributes.html) so that I don't have to surround the attribute in comments to make it a valid C construct. Note that before I was using C23 [attribute specifiers](https://en.cppreference.com/w/c/language/attributes), but that is still not widely supported, so had to write them as comments.

qmuntal:
I've upgraded the ubuntu CI agent to 22.04 as Ubuntu 20.04 is being sunsetted and in burnout period: https://github.com/actions/runner-images/issues/11101

gdams (approved):

shims.h
```diff
@@ -1,314 +1,380 @@
...
-// DEFINEFUNC_VARIADIC_3_0 acts like DEFINEFUNC but creates an alias with a more specific signature.
-// This is necessary to call variadic functions (functions that accept a variable number of arguments)
-// because variadic functions are not directly compatible with cgo. By defining a cgo-compatible alias
-// for each desired signature, the C compiler handles the variadic arguments rather than cgo.
-// Variadic functions are the only known incompatibility of this kind.
-// If you use this macro for a different reason, consider renaming it to something more general first.
-// See https://github.com/golang/go/issues/975.
-// The process is aborted if the function can't be loaded when using 3.0.0 or higher.
-//
+// The following includes are used be the headercheck tool.
```
dagood:
```suggestion
// The following includes are used by the checkheader tool.
```

(I agree this is enough doc for these includes.)

cmd/checkheader/main.go
```diff
@@ -93,212 +92,85 @@ func gccRun(program string) error {
...
 
-func tryConvertDirective(w io.Writer, l string) bool {
-	if strings.HasPrefix(l, "// #") {
-		fmt.Fprintln(w, l[len("// "):])
-		return true
+	for _, enum := range src.Enums {
+		if enum.Name == "_EVP_PKEY_OP_DERIVE" {
+			// This is defined differently in OpenSSL 3,
+			// but in our code it is only used in OpenSSL 1.
+			continue
```
dagood, line 119: I'm wondering if we should document this in `shims.h`. Having a `checkheader:ignore` line at least showed something odd was going on.

shims.h
```diff
@@ -1,314 +1,380 @@
...
+// - no tag: OpenSSL 1.0 or later
+// - legacy_1: Only OpenSSL 1
+// - 3: OpenSSL 3.0 or later
+// - 111: OpenSSL 1.1.1 or later
+
+// ERR API
+void ERR_error_string_n(unsigned long e, char *buf, size_t len);
+void ERR_clear_error(void);
+unsigned long ERR_get_error_line(const char **file, int *line) __attribute__((tag("legacy_1")));
+unsigned long ERR_get_error_all(const char **file, int *line, const char **func, const char **data, int *flags) __attribute__((tag("3")));
```
dagood, line 124: Hmm, could tidy up the syntax with `#define TAG(X) __attribute__((tag(X)))`, `#define VARIADIC(X) ...`, etc. (both `,` and multiple `__attribute__` seem to be valid.)

(Or really... `#define TAG(X)`, because we aren't asking gcc to parse these for us anyway. 😄)

cmd/mkcgo/main.go
```diff
@@ -0,0 +1,92 @@
...
+			log.Fatalf("failed to write unformatted source to file: %v", err)
+		}
+		log.Fatalf("for diagnosis, wrote unformatted source to %v", f)
+	}
+
+	// Write output. If no explicit output file is specified,
+	// // write both Go and C output to stdout.
+	if *fileName == "" {
+		for _, d := range [][]byte{data, cbuf.Bytes()} {
+			if _, err = os.Stdout.Write(d); err != nil {
```
dagood: Add a way to tell what file's being output? (txtar style?)

cmd/mkcgo/generate.go
```diff
@@ -0,0 +1,361 @@
...
+	}
+	return s
+}
+
+func retIsVoid(r *mkcgo.Return) bool {
+	return r.Type == "void"
+}
+
+// fnToGoParams returns source code for function f parameters.
+func fnToGoParams(fn *mkcgo.Func) string {
```
dagood, line 342: In theory these could also take `w io.Writer` to avoid string concat... but it seems like all of these operate on relatively small strings so it's likely ok.

cgo_go124.go
```diff
@@ -14,29 +14,29 @@ package openssl
...
-#cgo noescape go_openssl_EVP_PKEY_get_raw_private_key
-#cgo nocallback go_openssl_EVP_PKEY_get_raw_private_key
-#cgo noescape go_openssl_EVP_DigestSign
-#cgo nocallback go_openssl_EVP_DigestSign
-#cgo noescape go_openssl_EVP_Digest
-#cgo nocallback go_openssl_EVP_Digest
-#cgo noescape go_openssl_EVP_DigestUpdate
-#cgo nocallback go_openssl_EVP_DigestUpdate
+#cgo noescape RAND_bytes
+#cgo nocallback RAND_bytes
```
dagood, line 18: Maybe something to consider later, but could add a shims.h tag that generates these, to keep the info in one place?

goopenssl.c
```diff
@@ -55,8 +34,8 @@ go_openssl_fips_enabled(void* handle)
 
     // For OpenSSL 3.x.
     int (*EVP_default_properties_is_fips_enabled)(void*) = (int (*)(void*))dlsym(handle, "EVP_default_properties_is_fips_enabled");
-    void *(*EVP_MD_fetch)(void*, const char*, const char*) = (void* (*)(void*, const char*, const char*))dlsym(handle, "EVP_MD_fetch");
-    void (*EVP_MD_free)(void*) = (void (*)(void*))dlsym(handle, "EVP_MD_free");
+    _EVP_MD_PTR (*EVP_MD_fetch)(void*, const char*, const char*) = (_EVP_MD_PTR (*)(void*, const char*, const char*))dlsym(handle, "EVP_MD_fetch");
+    void (*EVP_MD_free)(_EVP_MD_PTR) = (void (*)(_EVP_MD_PTR))dlsym(handle, "EVP_MD_free");
```
dagood, line 38: Seeing these lines touched, it occurs to me that this small subset could potentially be marked by a tag and generated.

hash.go
```diff
@@ -2,7 +2,24 @@
...
-// #include "goopenssl.h"
+/*
+#include "shims.h"
+// go_hash_sum copies ctx into ctx2 and calls EVP_DigestFinal_ex using ctx2.
+// This is necessary because Go hash.Hash mandates that Sum has no effect
+// on the underlying stream. In particular it is OK to Sum, then Write more,
+// then Sum again, and the second Sum acts as if the first didn't happen.
+// It is written in C because Sum() tend to be in the hot path,
+// and doing one cgo call instead of two is a significant performance win.
+int EVP_MD_CTX_copy(_EVP_MD_CTX_PTR ctx, _EVP_MD_CTX_PTR ctx2);
```
dagood: Is this just a forward declaration? Or is it significant that it doesn't have the `const` and the names are different?
`int EVP_MD_CTX_copy(_EVP_MD_CTX_PTR out, const _EVP_MD_CTX_PTR in);`
(I'm rusty on how this works with C, or maybe it's cgo specific?)

hash.go
```diff
@@ -349,7 +366,7 @@ func (h *evpHash) BlockSize() int {
 func (h *evpHash) Sum(in []byte) []byte {
 	h.init()
 	out := make([]byte, h.Size(), maxHashSize) // explicit cap to allow stack allocation
-	if C.go_hash_sum(h.ctx, h.ctx2, base(out)) != 1 {
+	if C.go_hash_sum(h.ctx, h.ctx2, (*C.uchar)(unsafe.SliceData(out))) != 1 {
```
dagood, line 367: What is this change for?

internal/mkcgo/mkcgo.go
```diff
@@ -0,0 +1,67 @@
...
+	"slices"
+)
+
+// Source is a collection of type definitions and functions.
+type Source struct {
+	TypeDefs []*TypeDef
+	Enums    []*Enum
+	Funcs    []*Func
+	Files    []string
+	Comments []string // Directives starting with "// #"
```
dagood: From what I see, it has more than directives--maybe:

```suggestion
	Comments []string // All line comments. Directives in this slice start with "#"
```

internal/mkcgo/parse.go
```diff
@@ -0,0 +1,346 @@
...
+// parseFile parses file name and extracts all symbols.
+func (src *Source) parseFile(name string) error {
+	file, err := os.Open(name)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+
+	s := bufio.NewScanner(file)
+	var inBlockComment, inEnum bool
```
dagood, line 73: Scanning lines with a state machine (vs. e.g. scanning tokens with recursive descent) seems a little complicated. Specifically, passing a reference to the bool with `processComments(line, &inBlockComment)` is starting to make me wonder.

But... I'm not sure that it's really complicated enough to justify changing the approach. 😄 This isn't a general purpose tool, so it's not as if we need to be concerned about adding more complexity to it. If we e.g. hit an edge case, we can simply not do that.

openssl.go
```diff
@@ -253,14 +273,17 @@ func pbase(b []byte) unsafe.Pointer {
...
+	}
+	return unsafe.SliceData(b)
 }
 
 // sbase returns the address of the underlying array in b,
 // being careful not to panic when b has zero length.
-func sbase(b []byte) *C.char {
-	return (*C.char)(pbase(b))
+func sbase(b []byte) unsafe.Pointer {
+	return unsafe.Pointer(base(b))
```
dagood: I interpreted `sbase` as `signed base` before, now it doesn't seem to have a clear parallel with `base`?

rsa.go
```diff
@@ -278,8 +277,13 @@ func HashSignRSAPKCS1v15(priv *PrivateKeyRSA, h crypto.Hash, msg []byte) ([]byte
...
 func VerifyRSAPKCS1v15(pub *PublicKeyRSA, h crypto.Hash, hashed, sig []byte) error {
-	if pub.withKey(func(pkey C.GO_EVP_PKEY_PTR) C.int {
-		size := C.go_openssl_EVP_PKEY_get_size(pkey)
+	if pub.withKey(func(pkey _EVP_PKEY_PTR) int32 {
+		var size int32
+		if vMajor == 1 {
+			size = go_openssl_EVP_PKEY_size(pkey)
+		} else {
+			size = go_openssl_EVP_PKEY_get_size(pkey)
+		}
```
dagood: It seems like it would be less of a change if we added this mapping to the generated layer?

Also, this switch is done in more than one place: `evp.go` and `rsa.go`. That isn't *much*, but it's something.

dagood (approved):
I looked a bit deeper at the changes throughout the library, but also a few more ideas about the generator. It looks like it would be fine to merge and start using, but I have some questions and ideas.

cgo_go124.go
qmuntal: I like the idea, will do it in a follow up PR.

cmd/mkcgo/main.go
qmuntal: Done.

cmd/mkcgo/generate.go
qmuntal: These function returns strings because it makes it easier to use them in the caller site.

goopenssl.c
qmuntal: Fun that you say it. I have a WIP branch doing exactly that 😄.

hash.go
qmuntal: Good catch. This declarations are a leftover. Will remove.

hash.go
qmuntal: `base` no longer returns a `*C.char`. This is the only remaining place where we need to cast a slice to `*C.char`, so the case is don't in place.

openssl.go
qmuntal: Another leftover. `sbase` is not longer necessary, we already have `pbase` to get the `unsafe.Pointer` from a slice.

rsa.go
qmuntal: > It seems like it would be less of a change if we added this mapping to the generated layer?

Tricky, but done 😸.

PR #260, opened 2025-03-03
hkdf: Replace nil salt with a slice of a preallocated all zeros buffer

### LINKED ISSUE #253: TLS 1.3 HKDF-Extract with empty salt fails when using the KeyPair FIPS Provider
As instructed by @qmuntal I am opening this issue here in golang-fips/openssl.

The issue is that TLS 1.3 does not work in binaries built with microsoft/go and `-tags goexperiment.opensslcrypto` when using the KeyPair FIPS Provider for OpenSSL 3 (https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/4724).

I have been in contact with Keypair and the root cause is that Keypair FIPS provider enforces a minimum HMAC key length of 112 bits. However, when crypto/tls calls HKDF Extract() it sometimes does this with nil salt and consequently when that salt is used as the HMAC key, the FIPS provider's enforcement kicks in and triggers a panic.

The issue can be reproduced with this simple app:
```
package main

import (
	"crypto/tls"
	"fmt"
	"os"
)

func main() {
	if len(os.Args) <= 1 {
		panic("missing dst address")
	}

	c, err := tls.Dial("tcp", os.Args[1],
		&tls.Config{
			MinVersion: tls.VersionTLS13,
		})
	if err != nil {
		panic(err)
	}
	fmt.Printf("successfully connected to %s (%s)\n", os.Args[1], c.RemoteAddr().String())
	c.Close()
}
```

The panic looks following on microsoft/go1.23.6-20250211.6:
```
$ LD_LIBRARY_PATH=~/src/keypair/openssl-3.0.10-install/lib64 ./test www.google.com:443
panic: tls: HKDF-Extract invocation failed unexpectedly: EVP_PKEY_derive
        openssl error(s):

goroutine 1 [running]:
crypto/tls.(*cipherSuiteTLS13).extract(0x80b460?, {0x0?, 0x0?, 0x0?}, {0x0?, 0x0?, 0x0?})
        /usr/local/lib/microsoftgo/go1.23.6-20250211.6/src/crypto/tls/key_schedule.go:97 +0x193
crypto/tls.(*clientHandshakeStateTLS13).establishHandshakeKeys(0xc0001d9a08)
        /usr/local/lib/microsoftgo/go1.23.6-20250211.6/src/crypto/tls/handshake_client_tls13.go:514 +0x2ce
crypto/tls.(*clientHandshakeStateTLS13).handshake(0xc0001d9a08)
        /usr/local/lib/microsoftgo/go1.23.6-20250211.6/src/crypto/tls/handshake_client_tls13.go:132 +0x725
crypto/tls.(*Conn).clientHandshake(0xc000004708, {0x6af338, 0xc0000be140})
        /usr/local/lib/microsoftgo/go1.23.6-20250211.6/src/crypto/tls/handshake_client.go:375 +0x845
crypto/tls.(*Conn).handshakeContext(0xc000004708, {0x6af178, 0x833040})
        /usr/local/lib/microsoftgo/go1.23.6-20250211.6/src/crypto/tls/conn.go:1568 +0x3a6
crypto/tls.(*Conn).HandshakeContext(...)
        /usr/local/lib/microsoftgo/go1.23.6-20250211.6/src/crypto/tls/conn.go:1508
crypto/tls.dial({0x6af178?, 0x833040?}, 0xc000085e30, {0x645d46, 0x3}, {0x7ffc4884e085, 0x12}, 0xc00019a000)
        /usr/local/lib/microsoftgo/go1.23.6-20250211.6/src/crypto/tls/tls.go:159 +0x3a5
crypto/tls.DialWithDialer(...)
        /usr/local/lib/microsoftgo/go1.23.6-20250211.6/src/crypto/tls/tls.go:119
crypto/tls.Dial({0x645d46?, 0x100451089?}, {0x7ffc4884e085?, 0xc000085f40?}, 0x40f36b?)
        /usr/local/lib/microsoftgo/go1.23.6-20250211.6/src/crypto/tls/tls.go:173 +0x7a
main.main()
        /home/xxxxx/tls.go:14 +0x6f
```

Similar panic is triggered in microsoft/go1.24.0-20250212.5:
```
$ LD_LIBRARY_PATH=~/src/keypair/openssl-3.0.10-install/lib64 ./test www.google.com:443
panic: EVP_KDF_derive
        openssl error(s):

goroutine 1 [running]:
crypto/tls/internal/tls13.extract[...](0xc000290110?, {0x0?, 0xc00002f6b8?, 0x50d39b?}, {0x0?, 0x0?, 0x0?})
        /usr/local/lib/microsoftgo/go1.24.0-20250212.5/src/crypto/tls/internal/tls13/tls13.go:52 +0xbd
crypto/tls/internal/tls13.NewEarlySecret[...](0xc000290110, {0x0, 0x0?, 0x0?})
        /usr/local/lib/microsoftgo/go1.24.0-20250212.5/src/crypto/tls/internal/tls13/tls13.go:83 +0x35
crypto/tls.(*clientHandshakeStateTLS13).establishHandshakeKeys(0xc00002fa00)
        /usr/local/lib/microsoftgo/go1.24.0-20250212.5/src/crypto/tls/handshake_client_tls13.go:519 +0x2ac
crypto/tls.(*clientHandshakeStateTLS13).handshake(0xc00002fa00)
        /usr/local/lib/microsoftgo/go1.24.0-20250212.5/src/crypto/tls/handshake_client_tls13.go:134 +0x73e
crypto/tls.(*Conn).clientHandshake(0xc00028a008, {0x6e4b40, 0xc00028e000})
        /usr/local/lib/microsoftgo/go1.24.0-20250212.5/src/crypto/tls/handshake_client.go:379 +0x810
crypto/tls.(*Conn).handshakeContext(0xc00028a008, {0x6e4898, 0x8b2000})
        /usr/local/lib/microsoftgo/go1.24.0-20250212.5/src/crypto/tls/conn.go:1568 +0x39a
crypto/tls.(*Conn).HandshakeContext(...)
        /usr/local/lib/microsoftgo/go1.24.0-20250212.5/src/crypto/tls/conn.go:1508
crypto/tls.dial({0x6e4898?, 0x8b2000?}, 0xc00007be30, {0x686261, 0x3}, {0x7fff1e8f1085, 0x12}, 0xc000138000)
        /usr/local/lib/microsoftgo/go1.24.0-20250212.5/src/crypto/tls/tls.go:159 +0x3a5
crypto/tls.DialWithDialer(...)
        /usr/local/lib/microsoftgo/go1.24.0-20250212.5/src/crypto/tls/tls.go:119
crypto/tls.Dial({0x686261?, 0x457b49?}, {0x7fff1e8f1085?, 0xc00007bf40?}, 0x416cc8?)
        /usr/local/lib/microsoftgo/go1.24.0-20250212.5/src/crypto/tls/tls.go:173 +0x7a
main.main()
        /home/xxxxx/tls.go:14 +0x6f
```

According to RFC 5869 Section 2.2, the salt "if not provided, it is set to a string of HashLen zeros.".

Patching the OpenSSL HKDF bindings in the golang-fips/openssl/v2/hkdf.go to convert nil salt to zeroed buffer does fix the issue.

For microsoft/go1.23.6-20250211.6 the patch looks like following:
```
diff -ruN go1.23.6-20250211.6.orig/src/vendor/github.com/golang-fips/openssl/v2/hkdf.go go1.23.6-20250211.6/src/vendor/github.com/golang-fips/openssl/v2/hkdf.go
--- go1.23.6-20250211.6.orig/src/vendor/github.com/golang-fips/openssl/v2/hkdf.go       2025-02-21 11:58:44.705750417 +0200
+++ go1.23.6-20250211.6/src/vendor/github.com/golang-fips/openssl/v2/hkdf.go    2025-02-21 12:00:59.960965835 +0200
@@ -110,6 +110,11 @@
        if err != nil {
                return nil, err
        }
+
+       if salt == nil {
+               salt = make([]byte, h().Size())
+       }
+
        switch vMajor {
        case 3:
                if C.go_openssl_EVP_PKEY_CTX_set1_hkdf_key(c.ctx,
```

For microsoft/go1.24.0-20250212.5 the patch looks like following, though I have not done very extensive testing on go1.24:
```
diff -ruN go1.24.0-20250212.5.orig/src/vendor/github.com/golang-fips/openssl/v2/hkdf.go go1.24.0-20250212.5/src/vendor/github.com/golang-fips/openssl/v2/hkdf.go
--- go1.24.0-20250212.5.orig/src/vendor/github.com/golang-fips/openssl/v2/hkdf.go       2025-02-21 12:20:33.666148363 +0200
+++ go1.24.0-20250212.5/src/vendor/github.com/golang-fips/openssl/v2/hkdf.go    2025-02-21 12:22:19.831310298 +0200
@@ -119,6 +119,10 @@
                return nil, err
        }
 
+       if salt == nil {
+               salt = make([]byte, h().Size())
+       }
+
        switch vMajor {
        case 1:
                ctx, err := newHKDFCtx1(md, C.GO_EVP_KDF_HKDF_MODE_EXTRACT_ONLY, secret, salt, nil, nil)
```

samiponkanenssh:
hkdf: Replace nil salt with a slice of a preallocated all zeros buffer. This fixes HKDF when using KeyPair FIPS Provider for OpenSSL 3

Fixes golang-fips/openssl#253

### DIFF
--- a/hkdf.go
+++ b/hkdf.go
@@ -109,9 +109,14 @@ func (c *hkdf1) Read(p []byte) (int, error) {
 	return n, nil
 }
 
-// hkdfAllZerosSalt is a preallocated buffer of zeros used in ExtractHKDF()
-var hkdfAllZerosSalt [64]byte
+// hkdfAllZerosSalt is a preallocated buffer of zeros used in ExtractHKDF().
+// The size should be kept larger than the output length of any hash algorithm
+// used with HKDF.
+var hkdfAllZerosSalt [512]byte
 
+// ExtractHDKF implements the HDKF extract step.
+// If salt is nil, then this function replaces it internally with a buffer of
+// zeros whose length equals the output length of the specified hash algorithm.
 func ExtractHKDF(h func() hash.Hash, secret, salt []byte) ([]byte, error) {
 	if !SupportsHKDF() {
 		return nil, errUnsupportedVersion()
@@ -126,10 +131,10 @@ func ExtractHKDF(h func() hash.Hash, secret, salt []byte) ([]byte, error) {
 	// zeros, as specified in RFC 5896 and as OpenSSL EVP_KDF-HKDF documentation
 	// instructs. Take a slice of a preallocated buffer to avoid allocating new
 	// buffer per call.
-	if len(salt) == 0 {
+	if salt == nil {
 		hlen := h().Size()
 		if hlen > len(hkdfAllZerosSalt) {
-			return nil, errors.New("nil salt is not supported for the specified hash algorithm")
+			return nil, errors.New("nil salt is not supported: hash output length exceeds internal buffer size")
 		}
 		salt = hkdfAllZerosSalt[:hlen]
 	}
--- a/hkdf.go
+++ b/hkdf.go
@@ -110,9 +110,9 @@ func (c *hkdf1) Read(p []byte) (int, error) {
 }
 
 // hkdfAllZerosSalt is a preallocated buffer of zeros used in ExtractHKDF().
-// The size should be kept larger than the output length of any hash algorithm
+// The size should be kept as large as the output length of any hash algorithm
 // used with HKDF.
-var hkdfAllZerosSalt [512]byte
+var hkdfAllZerosSalt [64]byte
 
 // ExtractHDKF implements the HDKF extract step.
 // If salt is nil, then this function replaces it internally with a buffer of
@@ -130,13 +130,15 @@ func ExtractHKDF(h func() hash.Hash, secret, salt []byte) ([]byte, error) {
 	// If calling code specifies nil salt, replace it with a buffer of hashLen
 	// zeros, as specified in RFC 5896 and as OpenSSL EVP_KDF-HKDF documentation
 	// instructs. Take a slice of a preallocated buffer to avoid allocating new
-	// buffer per call.
+	// buffer per call, but fall back to allocating a buffer if preallocated
+	// buffer is not large enough.
 	if salt == nil {
 		hlen := h().Size()
 		if hlen > len(hkdfAllZerosSalt) {
-			return nil, errors.New("nil salt is not supported: hash output length exceeds internal buffer size")
+			salt = make([]byte, hlen)
+		} else {
+			salt = hkdfAllZerosSalt[:hlen]
 		}
-		salt = hkdfAllZerosSalt[:hlen]
 	}
 
 	switch vMajor {
--- a/hkdf.go
+++ b/hkdf.go
@@ -109,6 +109,12 @@ func (c *hkdf1) Read(p []byte) (int, error) {
 	return n, nil
 }
 
+// maxHKDFHashLength is maximum output length of any hash used with HKDF
+const maxHKDFHashLength = 64
+
+// hkdfAllZerosSalt is a preallocated buffer of zeros used in ExtractHKDF()
+var hkdfAllZerosSalt = make([]byte, maxHKDFHashLength)
+
 func ExtractHKDF(h func() hash.Hash, secret, salt []byte) ([]byte, error) {
 	if !SupportsHKDF() {
 		return nil, errUnsupportedVersion()
@@ -117,6 +123,18 @@ func ExtractHKDF(h func() hash.Hash, secret, salt []byte) ([]byte, error) {
 	md, err := hashFuncToMD(h)
 	if err != nil {
 		return nil, err
+	}
+
+	// If calling code specifies nil salt, replace it with a buffer of hashLen
+	// zeros, as specified in RFC 5896 and as OpenSSL EVP_KDF-HKDF documentation
+	// instructs. Take a slice of a preallocated buffer to avoid allocating new
+	// buffer per call.
+	if salt == nil {
+		hlen := h().Size()
+		if hlen > maxHKDFHashLength {
+			return nil, errors.New("nil salt is not supported for the specified hash algorithm")
+		}
+		salt = hkdfAllZerosSalt[:hlen]
 	}
 
 	switch vMajor {
--- a/hkdf.go
+++ b/hkdf.go
@@ -109,11 +109,8 @@ func (c *hkdf1) Read(p []byte) (int, error) {
 	return n, nil
 }
 
-// maxHKDFHashLength is maximum output length of any hash used with HKDF
-const maxHKDFHashLength = 64
-
 // hkdfAllZerosSalt is a preallocated buffer of zeros used in ExtractHKDF()
-var hkdfAllZerosSalt = make([]byte, maxHKDFHashLength)
+var hkdfAllZerosSalt [64]byte
 
 func ExtractHKDF(h func() hash.Hash, secret, salt []byte) ([]byte, error) {
 	if !SupportsHKDF() {
@@ -129,9 +126,9 @@ func ExtractHKDF(h func() hash.Hash, secret, salt []byte) ([]byte, error) {
 	// zeros, as specified in RFC 5896 and as OpenSSL EVP_KDF-HKDF documentation
 	// instructs. Take a slice of a preallocated buffer to avoid allocating new
 	// buffer per call.
-	if salt == nil {
+	if len(salt) == 0 {
 		hlen := h().Size()
-		if hlen > maxHKDFHashLength {
+		if hlen > len(hkdfAllZerosSalt) {
 			return nil, errors.New("nil salt is not supported for the specified hash algorithm")
 		}
 		salt = hkdfAllZerosSalt[:hlen]


hkdf.go
```diff
@@ -109,6 +109,12 @@ func (c *hkdf1) Read(p []byte) (int, error) {
 	return n, nil
 }
 
+// maxHKDFHashLength is maximum output length of any hash used with HKDF
+const maxHKDFHashLength = 64
+
+// hkdfAllZerosSalt is a preallocated buffer of zeros used in ExtractHKDF()
+var hkdfAllZerosSalt = make([]byte, maxHKDFHashLength)
```
qmuntal: It is simpler and more robust to use as array instead of an slice, like this `var hkdfAllZerosSalt [64]byte`. This way it can't be unintentionally shrank elsewhere, and we avoid defining two variables instead of one.

qmuntal (changes requested):

qmuntal (approved):
LGTM, thanks.

hkdf.go
```diff
@@ -119,6 +122,18 @@ func ExtractHKDF(h func() hash.Hash, secret, salt []byte) ([]byte, error) {
 		return nil, err
 	}
 
+	// If calling code specifies nil salt, replace it with a buffer of hashLen
```
dagood, line 130: Nitpicking, but does the caller really "specify nil salt" (or "specify no salt") or is it more accurate to say that the caller "doesn't specify a salt"? I'm not all that in tune with the wording around HKDF, but it seems more like "doesn't specify a salt" from what I read in https://github.com/golang-fips/openssl/issues/253.

Maybe the core of my confusion is: is a buffer of zeroes "no salt", or is it "salt of many zeroes" that we are injecting on the user's behalf when they don't have a preference? 😄

hkdf.go
```diff
@@ -119,6 +122,18 @@ func ExtractHKDF(h func() hash.Hash, secret, salt []byte) ([]byte, error) {
...
+	// zeros, as specified in RFC 5896 and as OpenSSL EVP_KDF-HKDF documentation
+	// instructs. Take a slice of a preallocated buffer to avoid allocating new
+	// buffer per call.
+	if len(salt) == 0 {
+		hlen := h().Size()
+		if hlen > len(hkdfAllZerosSalt) {
+			return nil, errors.New("nil salt is not supported for the specified hash algorithm")
```
dagood: This seems a little vague to me--the message seems to imply something is fundamentally incompatible, but "our internal buffer isn't long enough" doesn't seem fundamental. E.g. if someone wants to pass nil and use a hash function that requires a salt longer than 64 bytes, it would be ok for them to send a PR to golang-fips/openssl to increase the size of this buffer, right?

Also, "nil salt" could use an update. Here's an attempt:

```suggestion
			return nil, errors.New("caller requested a zero-filled buffer as salt by passing zero-length salt input, but the internal buffer is not large enough for the specified hash algorithm")
```

dagood: In theory, this could reduce compatibility with existing providers, right? If the provider supports empty salt, but golang-fips/openssl doesn't have enough buffer space, then returning this error means the call will fail when it could have succeeded if we were to fall back to passing an empty salt to the provider.

However, if we let the call continue rather than returning an error and the call *does* fail, then the user ends up with an unclear error message.

Alternative: we could handle this case safely by allocating memory if the buffer isn't large enough. It might be slower for a provider that could have handled an empty salt, but at least it isn't a crash. (We can then add tests that make sure that for all known hashes, it doesn't allocate, for futureproofness. (When we add a new compatible hash, it will "automatically" check that the buffer is long enough.))

I don't know if this is a practical concern in context.

hkdf.go
```diff
@@ -109,6 +109,9 @@ func (c *hkdf1) Read(p []byte) (int, error) {
 	return n, nil
 }
 
+// hkdfAllZerosSalt is a preallocated buffer of zeros used in ExtractHKDF()
+var hkdfAllZerosSalt [64]byte
+
 func ExtractHKDF(h func() hash.Hash, secret, salt []byte) ([]byte, error) {
```
dagood, line 120: Might be a good time to add a func comment, and describe the len 0 behavior in there?

hkdf.go
qmuntal: I initialli suggested to treat an empty salt as a nil salt, given that Go best practices say not to differentiate between empty and nil slices.

On the other hand, the Go crypto hkdf implementation does make this distinction:

https://github.com/golang/go/blob/0312e31ed197b3bf1434e8dbb283f0d2374d7457/src/crypto/internal/fips140/hkdf/hkdf.go#L16

Therefore we should do that to to keep compatibility.

hkdf.go
samiponkanenssh: I would just make the preallocated buffer larger, say 512 bytes.

hkdf.go
samiponkanenssh: Ok, reverting back the earlier change. Do you think the unit test should then use salt = nil instead of salt = []byte{}?

qmuntal: > Do you think the unit test should then use salt = nil instead of salt = []byte{}?

These tests come from the standard library, so I would rather keep them. If a provider doesn't support empty salts, then those tests will fail, and that's fine. It's on the provider to support an explicit empty salt.

hkdf.go
qmuntal: The largest hash size we support is 64, so I would stick to that. If `hlen` doesn't fit in the preallocated buffer, then it is fine to allocate, probably better than return an error.

hkdf.go
```diff
@@ -119,6 +127,18 @@ func ExtractHKDF(h func() hash.Hash, secret, salt []byte) ([]byte, error) {
...
+	// If calling code specifies nil salt, replace it with a buffer of hashLen
+	// zeros, as specified in RFC 5896 and as OpenSSL EVP_KDF-HKDF documentation
+	// instructs. Take a slice of a preallocated buffer to avoid allocating new
+	// buffer per call.
+	if salt == nil {
+		hlen := h().Size()
+		if hlen > len(hkdfAllZerosSalt) {
+			return nil, errors.New("nil salt is not supported: hash output length exceeds internal buffer size")
+		}
+		salt = hkdfAllZerosSalt[:hlen]
```
dagood: What I had in mind was keeping `hkdfAllZerosSalt` at 64 bytes (to fit the longest that we know can happen, for now) but handle this case gracefully:

```suggestion
		if hlen > len(hkdfAllZerosSalt) {
			salt = make([]byte, hlen)
		} else {
			salt = hkdfAllZerosSalt[:hlen]
		}
```

With that fallback in place, then at worst, if someone adds a new big hash to this repo later and doesn't update the size of `hkdfAllZerosSalt`, performance might be lower for that new hash. If the caller cares, they can even fix it themselves by passing in their own preallocated salt.

(A test that checks `h().Size() <= len(hkdfAllZerosSalt)` for all supported hashes would be nice, but IMO it's not critical for this PR to add it.)

qmuntal (approved):

dagood (approved):
Thanks!

PR #261, opened 2025-03-05
Implement version check functions in Go using special tags

qmuntal:
This PR removes all C code from `goopenssl.c` (the only remaining non-autogenerated C file!) and implement it in Go, with the main benefit of duplicating C's `go_openssl_fips_enabled` in favor of Go's `FIPS()`.

This is now possible thanks to have a more flexible mechanism to define and load OpenSSL function. Particularly, this PR added some new functions to `shims.h`, used when initializing the library, that were previously manually defined and loaded due to not having a valid library handle or not knowing the OpenSSL version at the time they were needed. We now can easily load functions on demand, and even mark them as optional, so implementing the backend initialization logic in Go is trivial.


init.go
```diff
@@ -2,56 +2,70 @@
...
-		return 0, 0, 0, errUnsupportedVersion()
+		return errUnsupportedVersion()
+	}
+
+	if initOnly {
+		switch vMajor {
+		case 1:
+			mkcgoLoad_init_1(handle)
+		case 3:
+			mkcgoLoad_init_3(handle)
```
dagood: Consider default case with `panic` in case this is updated poorly in the future, and to make it clearer to the reader what the intent is?

init.go
```diff
@@ -2,56 +2,70 @@
...
-)
+import "errors"
 
 // opensslInit loads and initialize OpenSSL.
 // If successful, it returns the major and minor OpenSSL version
 // as reported by the OpenSSL API.
 //
 // See Init() for details about file.
-func opensslInit(file string) (major, minor, patch uint, err error) {
+func opensslInit(file string, initOnly bool) error {
```
dagood: This naming doesn't seem ideal... based on the name, `opensslInit` clearly does init, but then what exactly does it mean to "only" init? The "extra" things it can optionally do are load some additional funcs and call `go_openssl_OPENSSL_init` and `go_openssl_OPENSSL_init_crypto`--which seem like initialization. 😄 

(Whenever I see a bool passed to a func I also question whether it should actually be two functions that share some logic to make call sites clear.)

Maybe a purpose-focused name? `opensslCapabilityCheckInit` vs. `opensslInit`?
Or `opensslInitShallow` vs. `opensslInit` to keep it generic?
(IMO at minimum, details in the doc comment.)

dagood (commented):
I like the approach, just a small naming and an edge case suggestion.

init.go
qmuntal: Done 👍

init.go
qmuntal: Agree and done 🐱.

init.go
```diff
@@ -67,7 +41,78 @@ func opensslInit(file string) (major, minor, patch uint, err error) {
...
+		return nil, err
+	}
+
+	switch vMajor {
+	case 1:
+		mkcgoLoad_init_1(handle)
+	case 3:
+		mkcgoLoad_init_3(handle)
+	default:
+		panic(errUnsupportedVersion())
```
dagood, line 80: Hmm, I think now that the other check for 1 or 3 is further away in `openLibrary`, this panic might actually make it a bit less obvious whether the caller will receive an error or a panic if the version isn't supported. An explicit comment might be reasonable:

```suggestion
		// We shouldn't get here: openLibrary should have already returned an error.
		panic(errUnsupportedVersion())
```

dagood (approved):

gdams (approved):

init.go
```diff
@@ -2,72 +2,146 @@
...
+	case 3:
+		mkcgoLoad_init_3(handle)
+	default:
+		// We shouldn't get here: openLibrary should have already returned an error.
+		panic(errUnsupportedVersion())
+	}
+	return func() {
+		dlclose(handle)
+		// Undo all the changes made in this function.
+		mkcgoUnload_version()
```
dagood, line 85: Could be a little more futureproof/self-evident if `openLibrary` returned a cancel func that gets called here and explicitly ignored in `opensslInit`.

Otherwise, arguably the comment should be one line up if we're talking about "in this function [and functions it called]", or one line down if "in this function [directly]".

PR #262, opened 2025-03-07
Move error handling to the autogenerated Go code

qmuntal:
This PR is a first step to solve #231. Now the Go autogenerated code is in charge of constructing a Go error from a OpenSSL error, which is still done using the `newOpenSSLError` function.

This PR shouldn't modify behavior, except for the few places we were previously omitting the error, but now we don't to make the code more robust.

The main benefit of this PR (other than helping fixing #231) is that it is no longer the developer responsibility to correctly construct a Go error when there is an OpenSSL failure. This was a problematic things, as demonstrated by the fact that there were many error handling patterns implemented across the codebase. Now the developer don't have to think about how to construct an error, just pass the autogenerated error up the chain.

### DIFF
--- a/evp.go
+++ b/evp.go
@@ -150,7 +150,7 @@ func loadHash(ch crypto.Hash) *hashAlgorithm {
 		// not created by EVP_MD_fetch has negative performance
 		// implications, as digest operations will have
 		// to fetch it on every call. Better to just fetch it once here.
-		md := go_openssl_EVP_MD_fetch(nil, go_openssl_EVP_MD_get0_name(hash.md), nil)
+		md, _ := go_openssl_EVP_MD_fetch(nil, go_openssl_EVP_MD_get0_name(hash.md), nil)
 		// Don't overwrite md in case it can't be fetched, as the md may still be used
 		// outside of EVP_MD_CTX, for example to sign and verify RSA signatures.
 		if md != nil {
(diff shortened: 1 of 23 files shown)


gdams (approved):

evp.go
```diff
@@ -221,12 +222,11 @@ func setupEVP(withKey withKeyFunc, padding int32,
...
 	var ctx _EVP_PKEY_CTX_PTR
-	withKey(func(pkey _EVP_PKEY_PTR) int32 {
-		ctx = go_openssl_EVP_PKEY_CTX_new(pkey, nil)
-		return 1
-	})
-	if ctx == nil {
-		return nil, newOpenSSLError("EVP_PKEY_CTX_new failed")
+	if _, err := withKey(func(pkey _EVP_PKEY_PTR) (int32, error) {
+		ctx, err = go_openssl_EVP_PKEY_CTX_new(pkey, nil)
+		return 0, err
```
dagood: It seems like withKey should just return an error rather than intentionally passing a placeholder (also in `rsa.go`). It looks like `int32, error` happens to line up with what the OpenSSL func returns naturally in a few cases, but not that many.

dagood (approved):

evp.go
qmuntal: Agree. Will do.

PR #263, opened 2025-03-08
Improve CheckVersion resource cleanup

qmuntal:
This is a follow up of #261, implementing what has been suggested in https://github.com/golang-fips/openssl/pull/261#discussion_r1985740702.

This PR also simplifies some related code and adds an outstanding missing `dlclose` call in the case the library loaded doesn't implement the OpenSSL version functions ([here](https://github.com/golang-fips/openssl/compare/init2?expand=1#diff-09d6aefefdc39400d949bac4606f951988cbe618740e96ecfee50236859da58bR127)).

### DIFF
--- a/init.go
+++ b/init.go
@@ -11,7 +11,7 @@ import (
 // osslHandle is the handle to the OpenSSL shared library loaded in the [Init] function.
 var osslHandle unsafe.Pointer
 
-// opensslInit loads and initialize OpenSSL..
+// opensslInit loads and initializes OpenSSL.
 //
 // See [Init] for details about file.
 func opensslInit(file string) error {
--- a/init.go
+++ b/init.go
@@ -51,9 +51,14 @@ func opensslInit(file string) error {
 
 // initForCheckVersion loads and initialize only the
 // functions required in [CheckVersion].
-// It returns a function that must be called to release the resources.
-// The function leaves all the global variables in the same state as they were
-// before the call.
+// It returns a close function that must be called to release the resources.
+//
+// This function modifies the vMajor, vMinor, and vPatch global variables as
+// well as other internal global variables that facilitate OpenSSL calls.
+//
+// If the function succeeds, calling the close function restores the previous
+// state of the global variables. If it fails, the global variables are restored
+// before returning.
 func initForCheckVersion(file string) (func(), error) {
 	prevMajor, prevMinor, prevPatch := vMajor, vMinor, vPatch
 	handle, close, err := openLibrary(file)
--- a/init.go
+++ b/init.go
@@ -18,31 +18,11 @@ var osslHandle unsafe.Pointer
 // See Init() for details about file.
 func opensslInit(file string) error {
 	// Load the OpenSSL shared library using dlopen.
-	handle, err := openLibrary(file)
+	handle, close, err := openLibrary(file)
 	if err != nil {
 		return err
 	}
 
-	loadOpenSSLFuncs(handle)
-
-	// Initialize OpenSSL.
-	go_openssl_OPENSSL_init()
-	if go_openssl_OPENSSL_init_crypto(
-		_OPENSSL_INIT_ADD_ALL_CIPHERS|
-			_OPENSSL_INIT_ADD_ALL_DIGESTS|
-			_OPENSSL_INIT_LOAD_CONFIG|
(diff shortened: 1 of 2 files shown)


init.go
```diff
@@ -18,12 +18,21 @@ var osslHandle unsafe.Pointer
 // See Init() for details about file.
```
dagood: Looks like the comment could use an update ("returns"). (Although it already needed one.)

gdams (approved):

init.go
```diff
@@ -33,84 +42,74 @@ func opensslInit(file string) error {
...
-		mkcgoLoad_111(handle)
-		mkcgoLoad_3(handle)
-	}
-}
-
 // initForCheckVersion loads and initialize only the
 // functions required in [CheckVersion].
 // It returns a function that must be called to release the resources.
 // The function leaves all the global variables in the same state as they were
 // before the call.
```
dagood: A bit more precise about the restore behavior:

```suggestion
// It returns a close function that must be called to release the resources.
//
// This function modifies the vMajor, vMinor, and vPatch global variables as
// well as other internal global variables that facilitate OpenSSL calls.
//
// If the function succeeds, calling the close function restores the previous
// state of the global variables. If it fails, the global variables are restored
// before returning.
```

init.go
```diff
@@ -33,84 +42,74 @@ func opensslInit(file string) error {
...
 func initForCheckVersion(file string) (func(), error) {
-	// This function can be called when the
 	prevMajor, prevMinor, prevPatch := vMajor, vMinor, vPatch
```
dagood, line 61: Not sure it's worth doing just for factoring's sake, but for readability, might be nice to have a name describing what's going on here:

```go
	prevMajor, prevMinor, prevPatch := vMajor, vMinor, vPatch
	restoreVersion := func() {
		vMajor, vMinor, vPatch = prevMajor, prevMinor, prevPatch
	}
```

(Maybe context from the func comment is plenty.)

init.go
```diff
@@ -33,84 +42,74 @@ func opensslInit(file string) error {
...
-		dlclose(handle)
-		// Undo all the changes made in this function.
-		mkcgoUnload_version()
-		mkcgoUnload_init_1()
-		mkcgoUnload_init_3()
-		vMajor, vMinor, vPatch = prevMajor, prevMinor, prevPatch
+		close()
 		if osslHandle != nil {
-			loadOpenSSLFuncs(osslHandle)
+			loadX(osslHandle)
```
dagood, line 95: Is this call being done (along with the version reset) to make it so that you can call `opensslInit` *then* `initForCheckVersion`, and `initForCheckVersion` will restore the state?

Is it a problem if we have OpenSSL 1 loaded, then call `initForCheckVersion` on OpenSSL 3, and then we end up calling `mkcgoLoad_init_3` on an OpenSSL 1 handle? I think that would result in an error for the non-optional bits.

qmuntal: d'oh, good catch. The failing test is probably related to that.

init.go
```diff
@@ -11,19 +11,26 @@ import (
 // osslHandle is the handle to the OpenSSL shared library loaded in the [Init] function.
 var osslHandle unsafe.Pointer
 
-// opensslInit loads and initialize OpenSSL.
-// If successful, it returns the major and minor OpenSSL version
-// as reported by the OpenSSL API.
+// opensslInit loads and initialize OpenSSL..
```
dagood:
```suggestion
// opensslInit loads and initializes OpenSSL.
```

dagood (approved):

qmuntal:
Hmm, tests are still flaky. Note that master is also flaky, but I thought this PR would fix it. I might have to partially revert #261.

qmuntal:
> Hmm, tests are still flaky. Note that master is also flaky, but I thought this PR would fix it. I might have to partially revert #261.

Found the culprit and fixed in https://github.com/golang-fips/openssl/pull/263/commits/7a5af55bf5504b7166c28110810154371d6b74cc. There was a race condition between `CheckVersion`, which is modifying `vMajor`, and the `opensslHMAC` finalizer, which is using `vMajor`.

The race was introduced some days ago in #261, as before that `CheckVersion` was not modifying any global state. This race shouldn't have happened in production, though, as `CheckVersion` is used to know which OpenSSL library to load, and it doesn't make much sense to call it after real OpenSSL work has been done.

PR #265, opened 2025-03-12
Move error handling to C code

### LINKED ISSUE #231: Error handling is not thead-safe

qmuntal:
This PR moves most of the error handling to the C side so that we don't have to worry about us retrieving OpenSSL errors from the wrong OpenSSL per-thread error queue (see #231).

The main trick is to autogenerate a new set of C wrappers, for those functions that return an error, with an additional error parameter that will be filled with error context if the function fails. The Go autogenerated functions will convert that error context into a Go error and return it to the caller.

This new error handling approach should have 0 performance impact when there are no errors. In the error situation, expect less allocations thanks to some small improvements I've done to the C ->Go error mapping process:

```
goos: windows
goarch: amd64
         │   old.txt   │               new.txt               │
         │   sec/op    │    sec/op     vs base               │
Error-12   5.035µ ± 1%   4.685µ ± 13%  -6.96% (p=0.019 n=10)

         │   old.txt    │               new.txt                │
         │     B/op     │     B/op      vs base                │
Error-12   2.184Ki ± 0%   1.537Ki ± 0%  -29.61% (p=0.000 n=10)

         │  old.txt   │              new.txt               │
         │ allocs/op  │ allocs/op   vs base                │
Error-12   20.00 ± 0%   10.00 ± 0%  -50.00% (p=0.000 n=10)
```

Note that this PR almost only changes `mkcgo` and autogenerated code, as #262 already did the job of refactoring the codebase to construct Go errors in the autogenerated code.

Fixes #231.


gdams (approved):

mertakman (approved):

cmd/mkcgo/generate.go
```diff
@@ -56,10 +69,13 @@ func generateGo(src *mkcgo.Source, w io.Writer) {
 		fmt.Fprintf(w, "}\n\n")
 	}
 
-	typedefs := make(map[string]string, len(src.TypeDefs))
-	for _, def := range src.TypeDefs {
-		typedefs[def.Name] = def.Type
-	}
+	// Generate error wrapper noescape function, which hides the
+	// error state pointer from the Go garbage collector.
```
dagood, line 50: Maybe link another use of this to make it obvious that it's a known way to do this and it's not something that Go will get wise to. 😄 (Could alternatively `linkname` it?)

```suggestion
	// error state pointer from the Go garbage collector.
	// An instance of https://github.com/golang/go/blob/d704ef76068eb7da15520b08dc7df98f45f85ffa/src/runtime/stubs.go#L194-L201
```

cmd/mkcgo/generate.go
```diff
@@ -492,3 +534,34 @@ func join(ps []*mkcgo.Param, fn func(int, *mkcgo.Param) string, sep string) stri
 	}
 	return strings.Join(params, sep)
 }
+
+const errStateNoEscapeName = "mkcgoNoEscape"
+const errStateStructName = "mkcgo_err_state"
```
dagood: Naming nit: at first glance these look like error values with that `err` prefix. Maybe `mkcgoErrState` for consistency, or `errorState`?

go_ossl_error.h
```diff
@@ -0,0 +1,78 @@
...
+			errs->code[i] = ERR_get_error_all(&file, &errs->line[i], NULL, NULL, NULL);
+		} else {
+			// OpenSSL 1 error handling
+			errs->code[i] = ERR_get_error_line(&file, &errs->line[i]);
+		}
+		if (errs->code[i] == 0) {
+			break;
+		}
+		if (file != NULL) {
+			// Copy the file name as the pointer we just retrieved will by OpenSSL
```
dagood: A few missing words? My guess:

```suggestion
			// Copy the file name as the pointer we just retrieved will be freed by OpenSSL
```

go_ossl_error.h
```diff
@@ -0,0 +1,78 @@
+#ifndef GO_OSSL_ERROR_H // only include this header once
+#define GO_OSSL_ERROR_H
+
+#include <shims.h>
+#include <stdlib.h> // for free()
+#include <string.h> // for strdup()
+
+int OPENSSL_version_major_Available();
```
dagood: It seems like at some point making a `zossl.h` file would make sense, or manually-maintained forward decls might build up. Maybe add a note in mkcgo about the reason for going with forward decls for future reference?

go_ossl_error.h
```diff
@@ -0,0 +1,78 @@
...
+	ERR_clear_error();
+}
+
+// mkcgo_err_retrieve retrieves the error state from OpenSSL.
+// It returns a pointer to a mkcgo_err_state structure
+// that contains the error codes, lines, and file names.
+// The caller is responsible for freeing the memory
+// by calling mkcgo_err_free.
+static mkcgo_err_state *mkcgo_err_retrieve() {
+	mkcgo_err_state *errs = (mkcgo_err_state *)malloc(sizeof(mkcgo_err_state));
```
dagood: Maybe use `calloc` to avoid the zeroing loop?

Or: use `malloc`, but remove the zeroing loop and instead make sure the single `for (int i = 0; i < ERR_NUM_MAX; i++) {` loop ends up assigning every element/field rather than `break`ing. (Or something similar.)

hash.go
```diff
@@ -374,8 +381,9 @@ func (h *evpHash) BlockSize() int {
 func (h *evpHash) Sum(in []byte) []byte {
 	h.init()
 	out := make([]byte, h.Size(), maxHashSize) // explicit cap to allow stack allocation
-	if C.go_hash_sum(h.ctx, h.ctx2, (*C.uchar)(unsafe.SliceData(out))) != 1 {
-		panic(newOpenSSLError("go_hash_sum"))
+	var _err *C.mkcgo_err_state
+	if C.go_hash_sum(h.ctx, h.ctx2, (*C.uchar)(unsafe.SliceData(out)), &_err) != 1 {
+		panic(newMkcgoErr("EVP_DigestFinal_ex", _err))
```
dagood: This might be an error for either `EVP_MD_CTX_copy` or `EVP_DigestFinal_ex`, right? Not sure how important it is for this particular string to be accurate, but maybe the more generic name is better to avoid potentially being misleading?

dagood: `errst` to match what `newMkcgoErr` calls it? `_` prefix seems unusual in hand-written Go.

```suggestion
	var errst *C.mkcgo_err_state
	if C.go_hash_sum(h.ctx, h.ctx2, (*C.uchar)(unsafe.SliceData(out)), &errst) != 1 {
		panic(newMkcgoErr("EVP_DigestFinal_ex", errst))
```

openssl.go
```diff
@@ -268,35 +269,50 @@ func base(b []byte) *byte {
...
+	for i := range 16 {
+		e := uint32(errst.code[i])
 		if e == 0 {
 			break
 		}
 		b.WriteByte('\n')
 		var buf [256]byte
 		go_openssl_ERR_error_string_n(e, base(buf[:]), len(buf))
-		b.WriteString(string(buf[:]) + "\n\t" + C.GoString((*C.char)(unsafe.Pointer(file))) + ":" + strconv.Itoa(int(line)))
+		b.Write(buf[:])
```
dagood: Maybe fix null terminator handling in buf while here?

```suggestion
		if termIdx := bytes.IndexByte(buf[:], 0); termIdx != -1 {
			b.Write(buf[:termIdx])
		} else {
			b.Write(buf[:])
		}
```

(Brief mention in https://github.com/golang-fips/openssl/issues/266.)

openssl_test.go
```diff
@@ -146,3 +147,45 @@ func TestFIPSCapable(t *testing.T) {
...
+	for range 10 {
+		wg.Add(2)
+		go func() {
+			defer wg.Done()
+			_, _, _, _, _, _, _, _, err := openssl.GenerateKeyRSA(1)
+			if err == nil {
+				t.Error("expected error, got nil")
+			} else if !strings.Contains(err.Error(), "key size too small") {
+				t.Errorf("expected error to contain 'rsa routines', got %q", err)
+			}
```
dagood, line 168: Maybe also check that `err.Error()` doesn't contain null bytes, for easy coverage?

dagood (approved):
Nice! A small handful of minor suggestions.

cmd/mkcgo/generate.go
qmuntal: > (Could alternatively linkname it?)

The `linkname` directive is locked down, better avoid it: https://github.com/golang/go/issues/67401.

cmd/mkcgo/generate.go
qmuntal: Good point. Will do.

go_ossl_error.h
qmuntal: `calloc` 👍

openssl_test.go
qmuntal: Good call. Done.

go_ossl_error.h
qmuntal: Good point. I've updated `mkcgo` to also generate `zossl.h`, which also enabled other simplifications.

hash.go
qmuntal: Fixed.

dagood (approved):

PR #268, opened 2025-03-20
Add custom prefix to autogenerated C functions

qmuntal:
In #259 I refactored the C bindings to autogenerate them. While doing that, I removed the `go_openssl` prefix from them, leaving the C function names with the same name as defined in the OpenSSL headers.

The problem is that when this Go module is compiled as a shared library, dynamically loaded in another module, and that module ends up dynamically loading the real OpenSSL (which is the case of all the OpenSSL providers) then the OpenSSL functions that we defined are loaded twice, from us and from OpenSSL, causing undefined behavior.

To avoid this issue, we better add a custom prefix to the autogenerated C functions. This PR does that by applying the following changes:

- Autogenerated C functions now are prefixed with `_mkcgo_`.
- All autogenerated C functions are now declared in `zossl.h`, not just the one needing an error wrapper, as `shims.h` no longer contains real function names (unless OpenSSL is dynamically loaded).
- Replaced `shims.h` for `zoosl.h` in those fails that were including it.
- Bonus: We were previously generating two wrapper functions for those C functions that required error handling, but only the one that did the error handling was really called from Go. I've removed the non-error variant.


gdams (approved):

cmd/mkcgo/generate.go
```diff
@@ -585,12 +586,33 @@ func fnCErrWrapperParams(fn *mkcgo.Func, addName bool) string {
 	return args
 }
 
-// fnCErrWrapperName returns the name of the error wrapper function for function f.
-func fnCErrWrapperName(fn *mkcgo.Func) string {
-	return "_mkcgo_err_" + fn.CName
+// fnGoName returns the Go function name for function f.
+func fnGoName(fn *mkcgo.Func) string {
+	// TODO: use a prefix that is not OpenSSL specific.
```
dagood, line 591: TODO, open an issue or fix?

dagood (approved):

cmd/mkcgo/generate.go
qmuntal: Fixed in https://github.com/golang-fips/openssl/pull/269 😄

PR #269, opened 2025-03-20
Move OpenSSL bindings to a dedicated package

qmuntal:
The `openssl` package is starting to have to many symbols in it, even when most as autogenerated. This PR moves all the autogenerated code to `internal/ossl`, making it easier to differentiate autogenerated symbols and helping gopls do its job faster.

There is no new logic here, just good old `CTRL+C` + `CTRL+V`.


internal/ossl/ossl.go
```diff
@@ -0,0 +1,39 @@
+package ossl
```
dagood, line 2: Package doc comment?

cmd/mkcgo/generate.go
```diff
@@ -616,3 +610,20 @@ func fnNeedErrWrapper(fn *mkcgo.Func) bool {
...
+		return ""
+	}
+	if *private {
+		return strings.ToLower(name[0:1]) + name[1:]
+	}
+	if name[0] == '_' {
+		name = name[1:]
+	}
+	if name == "" {
+		return ""
```
dagood: Is `""` actually used? It might be better to `panic` if someone tries to translate a `_` to Go (or passes `""`, for that matter), because I imagine omitting a symbol in the generated code might cause some weird syntax errors downstream.

cmd/mkcgo/main.go
```diff
@@ -16,6 +16,7 @@ var (
 	fileName      = flag.String("out", "", "output file name (standard output if omitted)")
 	includeHeader = flag.String("include", "", "include header file")
 	packageName   = flag.String("package", "", "package name")
+	private       = flag.Bool("private", false, "all Go generated symbols are kept unexported")
```
dagood, line 19: Is this exercised? Maybe at least have a test case to make sure the feature keeps working. (And maybe document when it'd be used?)

cmd/mkcgo/generate.go
```diff
@@ -616,3 +610,20 @@ func fnNeedErrWrapper(fn *mkcgo.Func) bool {
...
 	return !fn.Variadic() // cgo doesn't support variadic functions
 }
+
+// goSymName returns the Go symbol name for a C symbol name.
+func goSymName(name string) string {
+	if name == "" {
+		return ""
+	}
+	if *private {
+		return strings.ToLower(name[0:1]) + name[1:]
```
dagood: Hmm, is this func used in a place where `EVP_aes_128_gcm` would become `eVP_aes_128_gcm`? A lowercase prefix might be better to be consistent. (Keeping the case the same as what OpenSSL has in docs seems like it helps readability too, to me.)

dagood: I suppose the same could be said the other way, although there's only `point_conversion_form_t` in that direction. 🤷

dagood (approved):
LGTM, just minor suggestions.

cmd/mkcgo/main.go
qmuntal: I use this for other experiments. Plan to add some test cases in a follow-up PR.

gdams (approved):

PR #270, opened 2025-03-21
Add tests for `internal/mkcgo`

qmuntal:
This PR adds unit tests for the `internal/mkcgo` package. The coverage is 100% (if that means somethinf 😄):

```
go test -cover ./internal/mkcgo 
ok      github.com/golang-fips/openssl/v2/internal/mkcgo        7.642s  coverage: 99.6% of statements
```

I've refactored some APIs to make to make them easier to test, added some additional validations, and fixed some bugs (which `shims.h` was not exercising) along the way.

Test for `cmd/mkcgo` will come in a follow-up PR.

### DIFF
--- a/internal/mkcgo/mkcgo.go
+++ b/internal/mkcgo/mkcgo.go
@@ -1,7 +1,12 @@
+// The mkcgo package provides a C header syntax parser.
+// It supports just the necessary features to parse the OpenSSL symbols used in this project.
 package mkcgo
 
 import (
+	"errors"
+	"fmt"
 	"slices"
+	"strings"
 )
 
 // Source is a collection of type definitions and functions.
@@ -9,9 +14,10 @@ type Source struct {
 	TypeDefs []*TypeDef
 	Enums    []*Enum
 	Funcs    []*Func
-	Files    []string
-	Comments []string // All line comments. Directives in this slice start with "#"
+	Comments []string // All line comments. Leading and trailing spaces are trimmed.
 	Includes []string // All #include directives, without the #include prefix.
+
+	symbols map[string]struct{} // All symbols defined in the source.
 }
 
 // TypeDef describes a type definition.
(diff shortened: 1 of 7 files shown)


internal/mkcgo/mkcgo.go
```diff
@@ -26,18 +32,63 @@ type Enum struct {
...
+			bld.WriteString(", ")
+		}
+		bld.WriteString(p.Type)
+		if p.Name != "" {
+			bld.WriteByte(' ')
+			bld.WriteString(p.Name)
+		}
+	}
+	bld.WriteString(") ")
+	bld.WriteString(fmt.Sprintf("%v", f.FuncAttrs))
```
dagood: IMO the result of this is puzzling... `int TestFunc() {[]  false false  false false}` in the test case. Maybe `FuncAttrs` should have a `String()` too, so it can output nothing in this case, and so it doesn't rely on positional bools in other cases?

internal/mkcgo/mkcgo_test.go
```diff
@@ -0,0 +1,402 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			fn := &mkcgo.Func{Params: tt.params}
+			if got := fn.Variadic(); got != tt.expected {
+				t.Errorf("Func.Variadic() = %v, want %v", got, tt.expected)
+			}
+		})
+	}
+}
+func TestSourceTags(t *testing.T) {
```
dagood, line 60:
```suggestion

func TestSourceTags(t *testing.T) {
```

internal/mkcgo/mkcgo_test.go
```diff
@@ -0,0 +1,402 @@
...
+			want:    `invalid signature`,
+		}, {
+			content: `void (void);`,
+			want:    `missing name`,
+		}, {
+			content: `void foo(,);`,
+			want:    `empty parameter`,
+		}, {
+			content: `void foo2(int a) __attribute__((variadic("foo")));`,
+			want:    "variadic target not added yet",
```
dagood: This makes me realize "yet" can be misleading--it can imply that the parser knows that the target *will* be added later. But in this case, it won't ever be added.

```suggestion
			want:    "variadic target not found in preceding code",
```

(Or: could remove the order dependency by doing a second pass to link up the targets. Normally I'd say this is simply better, but we're in C land, so order dependency is something the user should be expecting. 😄)

internal/mkcgo/mkcgo_test.go
```diff
@@ -0,0 +1,402 @@
...
+			want:    `can't extract function attributes: __attribute__ should be followed by double parentheses`,
+		}, {
+			content: `void foo(void) __attribute__((tag("a")optional));`,
+			want:    `can't extract function attributes: parameters must be separated by commas`,
+		}, {
+			content: `void foo(void) __attribute__((tag("a"());`,
+			want:    `can't extract function attributes: unbalanced parentheses`,
+		}, {
+			content: `void foo(void) __attribute__(());`,
+			want:    `can't extract function attributes: empty attribute`,
```
dagood: An empty attribute list is valid, according to https://gcc.gnu.org/onlinedocs/gcc/Attribute-Syntax.html.

dagood (approved):

gdams (approved):

PR #272, opened 2025-04-14
Add support for the TLS13-KDF algorithm

nicholasberlin:
We are attempting to run a go binary on Ubuntu with FIPS enabled. But, the openssl shared object is rejecting our binary's usage of TLS1.3 because the `iv` length. 

https://github.com/microsoft/go/issues/1626

~This PR makes use of the KDF specialization carve out for TLS1.3 when the code detects it.~
This PR adds functions for external users to make use of TLS13-KDF


qmuntal (changes requested):
Thanks for submitting this. I would prefer to have a dedicated on-shot API for TLS 1.3 KDF. Something like `func SupportsTLS13KDF()` and `func ExpandTLS13KDF(fh func() hash.Hash, secret []byte, label string, context []byte, keyLength int) []byte`. The caller can decide whether to use this or not based on `SupportsTLS13KDF`.

hkdf.go
```diff
@@ -27,6 +28,18 @@ func SupportsHKDF() bool {
 	}
 }
 
+// SupprtsTLS13KDF reports whether the current OpenSSL version supports TLS13-KDF.
+func SupportsTLS13KDF() bool {
+	switch vMajor {
+	case 3:
```
dagood, line 35: This needs a `case 1` so a caller using OpenSSL 1 can get a `false` rather than a panic.

hkdf.go
```diff
@@ -214,6 +227,36 @@ func ExpandHKDFOneShot(h func() hash.Hash, pseudorandomKey, info []byte, keyLeng
 	return out, nil
 }
 
+// ExpandHKDFOneShotTLS13_KDF_Support derives a key from the given hash, key, and context info, and will use
+// the TLS 1.3 KDF if the info parameter is in the expected format.
+func ExpandHKDFOneShotTLS13_KDF_Support(h func() hash.Hash, pseudorandomKey, info []byte, keyLength int) ([]byte, error) {
```
dagood: Why is "Support" part of this func's name?

Also, this module only uses `_` in an exported name to separate two numbers so far, I'm not sure it's worth breaking the convention here.

```suggestion
// ExpandHKDFOneShotTLS13KDF derives a key from the given hash, key, and context info, and will use
// the TLS 1.3 KDF if the info parameter is in the expected format.
func ExpandHKDFOneShotTLS13KDF(h func() hash.Hash, pseudorandomKey, info []byte, keyLength int) ([]byte, error) {
```

hkdf.go
```diff
@@ -261,6 +304,155 @@ func (c *hkdf3) finalize() {
...
+// This function parses the info parameter for TLS 1.3 KDF.
+// The info parameter is expected to be in the format:
+//
+//    +--------------+-----------+-------------------+------------+------------------+
+//    | total length | labelLen  | label bytes...    | contextLen | context bytes... |
+//    | 2 bytes      | 1 byte    | labelLen bytes    | 1 byte     | contextLen bytes |
+//    +--------------+-----------+-------------------+------------+------------------+
+//
+// The label bytes are expected to be begin with "tls13 ".
+//
```
dagood: Comment doesn't seem complete--my suggestions would be rearrange, add info about returns, fix small typo, describe links:

```suggestion
// ParseForTLS13 parses the info parameter for TLS 1.3 KDF.
// The info parameter is expected to be in the format:
//
//    +--------------+-----------+-------------------+------------+------------------+
//    | total length | labelLen  | label bytes...    | contextLen | context bytes... |
//    | 2 bytes      | 1 byte    | labelLen bytes    | 1 byte     | contextLen bytes |
//    +--------------+-----------+-------------------+------------+------------------+
//
// The label bytes are expected to begin with "tls13 ".
//
// Based on https://datatracker.ietf.org/doc/html/rfc8446#section-7.1
// RFC8446 mentioned by https://docs.openssl.org/3.4/man7/EVP_KDF-TLS13_KDF/
//
// If info matches expectations, isTLS13 is true and label and context are returned.
// Otherwise, isTLS13 is false.
```

A few other high-level questions:

* Could this use a more typical `ok` (or `err`), where it's the final returned value, to make it more intuitive to use by convention?
* Are naked returns really necessary? IMO they make it harder to follow otherwise simple logic. (And often projects forbid them in nontrivial funcs, enforced by linters.)
* Use e.g. `info = info[1:]` rather than introducing a new variable, to simplify?
* Why is this exported?

hkdf.go
```diff
@@ -261,6 +304,155 @@ func (c *hkdf3) finalize() {
...
+	// Success, set the out parameters
+	label = labelBytes[len("tls13 "):]
+	context = info[cursor : cursor+contextLen]
+	isTLS13 = true
+
+	return
+}
+
+// fetchHKDF3_TLS13KDF fetches the "TLS13-KDF" for TLS 1.3 handshakes, and otherwise falls back to "HKDF".
+func newHKDFCtx3_with_TLS13_KDF_Support(md ossl.EVP_MD_PTR, mode int32, secret, salt, pseudorandomKey, info []byte) (_ ossl.EVP_KDF_CTX_PTR, err error) {
```
dagood: Name mismatch

dagood (commented):
Mostly localized style comments, please feel free to wait for @qmuntal's next review to see if  they're applicable in the big picture. 🙂

hkdf.go
nicholasberlin: > Why is this exported?

So that it can be unit tested. That ok? Should I move hkdf_test.go to the `openssl` package?

dagood: > > Why is this exported?
> 
> So that it can be unit tested. That ok? Should I move hkdf_test.go to the `openssl` package?

I think what I would do here is move this func to a new `internal` package. This way it can't be unintentionally used by external importers, but it can still be tested directly without doing something unusual in the `openssl` package. (Admittedly, there isn't precedent for this in this repo.)

A few other options:

* Add a new test file for these private unit tests that uses `package openssl`. (Without touching `hkdf_test.go`.)
* Add an exporter `func` to `export_test.go`.
* Change the test to only go through the intended-public APIs and make inferences based on the results.

@qmuntal probably has a firmer opinion about this. 😄

qmuntal: We don't need this function if `newHKDFCtx3_with_TLS13_KDF_Support` accepts a label and a context as byte slices. I don't see the benefit on passing them as a tlf1.3 info. This way, when this KDF is available, callers can avoid encoding the info and just rely on the the kdf to return the correctly encoded data.

hkdf.go
qmuntal: See https://github.com/golang-fips/openssl/pull/272#discussion_r2052870493.

Also, `HKDF` should not appear in the name of the function.

nicholasberlin:
> We don't need this function if newHKDFCtx3_with_TLS13_KDF_Support accepts a label and a context as byte slices. I don't see the benefit on passing them as a tlf1.3 info. This way, when this KDF is available, callers can avoid encoding the info and just rely on the the kdf to return the correctly encoded data.

I believe I've addressed everything up to this comment. Back to it tomorrow.

nicholasberlin:
Oh, and this one:
>Also, HKDF should not appear in the name of the function.

hkdf.go
nicholasberlin: Ok, so I understand ...

In order to avoid creating/parsing `info`, [`tls13.ExpandLabel`](https://github.com/microsoft/go/blob/microsoft/main/patches/0004-Use-crypto-backends.patch#L2456) will effectively need to use the new functions of this PR to check support for and use `TLS13-KDF`. (Leaving the original code as a fallback)

This will involve adding stubs for `SupportsTLS13KDF` and `ExpandTLS13KDF` to each backend in `crypto/internal/backend`. Similar to [this](https://github.com/microsoft/go/blob/microsoft/main/patches/0003-Implement-crypto-internal-backend.patch#L2521)

And that will further mean importing `boring "crypto/internal/backend"` in `crypto/tls/internal/tls13/tls13.go`. Similar to [this](https://github.com/microsoft/go/blob/microsoft/main/patches/0004-Use-crypto-backends.patch#L1128).

Make sense?

qmuntal: Yep, that's exactly what I had in my mind.

nicholasberlin:
Ok, I think this is ready for another round

.gitignore
```diff
@@ -1,2 +1,3 @@
 **/.DS_Store
 .vscode/
+testdata
```
qmuntal: `testdata` shouldn't be ignored.

hkdf.go
qmuntal: Still not resolved.

hkdf.go
```diff
@@ -261,6 +298,63 @@ func (c *hkdf3) finalize() {
...
+	checkMajorVersion(3)
+
+	kdf, err := ossl.EVP_KDF_fetch(nil, _OSSL_KDF_NAME_TLS13_KDF.ptr(), nil)
+	if err != nil {
+		return nil, err
+	}
+	return kdf, nil
+})
+
+// newTLS13KDFExpandCtx3 fetches the "TLS13-KDF" for TLS 1.3 handshakes, and otherwise falls back to "HKDF".
```
qmuntal: There is no fall back here.

```suggestion
// newTLS13KDFExpandCtx3 fetches the "TLS13-KDF" for TLS 1.3 handshakes.
```

hkdf.go
```diff
@@ -214,6 +226,31 @@ func ExpandHKDFOneShot(h func() hash.Hash, pseudorandomKey, info []byte, keyLeng
...
+	}
+
+	md, err := hashFuncToMD(h)
+	if err != nil {
+		return nil, err
+	}
+
+	out := make([]byte, keyLength)
+
+	ctx, err := newTLS13KDFExpandCtx3(md, []byte(label), context, pseudorandomKey)
```
qmuntal: `label` is already a `[]byte`.

```suggestion
	ctx, err := newTLS13KDFExpandCtx3(md, label, context, pseudorandomKey)
```

qmuntal (changes requested):

hkdf.go
nicholasberlin: @dagood, I updated the `default` case to return `false` instead of calling `panic`. Does that resolve your concern, or would you prefer an explicit `case 1`?

nicholasberlin: meh, I just added an explicit case.

.gitignore
nicholasberlin: ok, removed ... but why?

hkdf.go
nicholasberlin: thanks.

hkdf.go
nicholasberlin: thanks.

hkdf.go
nicholasberlin: Removed `HKDF` here: 26ab78edfb2abeb7bc171caa2f6a4dcc1e03198b

hkdf.go
qmuntal: > I updated the default case to return false instead of calling panic. Does that resolve your concern, or would you prefer an explicit case 1?

Oh, I didn't notice that change. Adding the `case 1: return false` is the right thing to do. Then the `default` case should panic with `panic(errUnsupportedVersion())`, like it is done in other places.

hkdf.go
nicholasberlin: > Yep, that's exactly what I had in my mind.

This happened here: https://github.com/golang-fips/openssl/commit/26ab78edfb2abeb7bc171caa2f6a4dcc1e03198b

hkdf.go
nicholasberlin: ah, makes sense: [bfd2f2e](https://github.com/golang-fips/openssl/pull/272/commits/bfd2f2ed082f85fdded7497aecf49812c0ce40f7)

.gitignore
qmuntal: Well, you are not using `testdata` anywhere in this PR, so I don't see the need to add it to `.gitignore`. Also, `testdata` is commonly used to store test fixtures that are used when running tests, so you don't want to ignore them.

nicholasberlin: Oh, `testdata` was an artifact of the fuzzing test for the parsing function, which is gone now.

qmuntal (approved):
LGTM, thanks!

dagood (approved):

PR #278, opened 2025-05-15
Support serializing SymCrypt hash objects

qmuntal:
OpenSSL does not currently support resumable hashing, where an in-progress hash operation's state is saved, and can be loaded later to finish the hash.

We do support serializing the hash state of the built-in OpenSSL providers, given that we can get a pointer to it and also we know it's memory layout. That doesn't work for other providers, though.

The SymCrypt provider is adding some custom parameters that allows getting and setting the internal hash object state. See https://github.com/microsoft/SymCrypt-OpenSSL/pull/117.

This PR uses those new parameters to implement the hash serialization methods when using the SymCrypt provider.

For https://github.com/microsoft/go/issues/1671.


gdams (approved):

hash.go
```diff
@@ -15,6 +15,22 @@ import (
...
+	magicMD5     = "md5\x01"
+	magic1       = "sha\x01"
+	magic224     = "sha\x02"
+	magic256     = "sha\x03"
+	magic384     = "sha\x04"
+	magic512_224 = "sha\x05"
+	magic512_256 = "sha\x06"
+	magic512     = "sha\x07"
+
+	marshaledSizeMD5 = len(magicMD5) + 4*4 + 64 + 8
```
dagood: Maybe worth a short comment describing the meaning of each add term?

providersymcrypt.go
```diff
@@ -0,0 +1,374 @@
...
+}
+
+// swapEndianessInt32 swaps the endianness of the given byte slice
+// in place. It assumes the slice is a backup of a 32-bit integer array.
+func swapEndianessInt32(d []uint8) {
+	for i := 0; i < len(d); i += 4 {
+		d[i], d[i+3] = d[i+3], d[i]
+		d[i+1], d[i+2] = d[i+2], d[i+1]
+	}
+
```
dagood, line 114:
```suggestion
```

providersymcrypt.go
```diff
@@ -0,0 +1,374 @@
...
+	d = d[len(chain):]
+	copy(buffer[:], d)
+	d = d[len(buffer):]
+	_, length := consumeUint64(d)
+	return newUINT64(length)
+}
+
+// swapEndianessInt32 swaps the endianness of the given byte slice
+// in place. It assumes the slice is a backup of a 32-bit integer array.
+func swapEndianessInt32(d []uint8) {
```
dagood: Calling it `Int32` rather than `Uint32` might be confusing later? AFAIK it doesn't matter, but in that case, I figure it 
 might as well match the rest of the file.

dagood (approved):
LGTM, just doc/style nits. I like the factoring, was fairly straightforward to follow along.

PR #287, opened 2025-07-24
Wrap hash state not marshallable errors with ErrUnsupported

mertakman:
This pull request refactors error handling in the OpenSSL package to provide easier to assert error messages by wrapping errors with additional context using errors.ErrUnsupported. It also includes minor import adjustments in the affected files.
This PR addresses one of the steps of: https://github.com/microsoft/go/issues/1683

### DIFF
--- a/hash.go
+++ b/hash.go
@@ -6,7 +6,6 @@ import "C"
 import (
 	"crypto"
 	"errors"
-	"fmt"
 	"hash"
 	"runtime"
 	"strconv"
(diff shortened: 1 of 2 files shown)


hash.go
```diff
@@ -6,6 +6,7 @@ import "C"
 import (
 	"crypto"
 	"errors"
+	"fmt"
```
qmuntal: Can't use `fmt` in this package, as it is not allowed as a dependency of `crypto` packages.

Same for the CNG and Darwin backends.

qmuntal (changes requested):

qmuntal (approved):

gdams (approved):

PR #299, opened 2025-10-17
CI: bump actions and add caching to openssl build

gdams:
Currently the openssl build takes 2-3 minutes for each job, this can be significantly reduced with caching


qmuntal (approved):
Nice!

gdams:
hmm I'm going to close this for now, the restore cache action doesn't work with elevated permissions and the workaround is a lot of extra complexity in the code

PR #310, opened 2025-10-31
implement mlkem for OpenSSL

gdams:


mlkem.go
```diff
@@ -0,0 +1,374 @@
...
+
+// DecapsulationKeyMLKEM768 is the secret key used to decapsulate a shared key
+// from a ciphertext. It includes various precomputed values.
+type DecapsulationKeyMLKEM768 [seedSizeMLKEM]byte
+
+// GenerateKeyMLKEM768 generates a new decapsulation key, drawing random bytes from
+// the default crypto/rand source. The decapsulation key must be kept secret.
+func GenerateKeyMLKEM768() (DecapsulationKeyMLKEM768, error) {
+	var dk DecapsulationKeyMLKEM768
+	_, err := rand.Read(dk[:])
```
qmuntal: This doesn't seem right. We should use OpenSSL to generate the mlkem key, not our own bespoken thing, even if it is a one-liner.

mlkem.go
```diff
@@ -0,0 +1,374 @@
...
+//
+// The shared key must be kept secret.
+func (dk DecapsulationKeyMLKEM768) Decapsulate(ciphertext []byte) (sharedKey []byte, err error) {
+	return performDecapsulation(ossl.NID_ML_KEM_768, dk[:], ciphertext)
+}
+
+// EncapsulationKey returns the public encapsulation key necessary to produce
+// ciphertexts.
+func (dk DecapsulationKeyMLKEM768) EncapsulationKey() EncapsulationKeyMLKEM768 {
+	pkey, err := createMLKEMPrivateKey(ossl.NID_ML_KEM_768, dk[:])
```
qmuntal: This function is almost the same as `DecapsulationKeyMLKEM768.EncapsulationKey`, factor it out.

mlkem.go
```diff
@@ -0,0 +1,374 @@
...
+
+// DecapsulationKeyMLKEM1024 is the secret key used to decapsulate a shared key
+// from a ciphertext. It includes various precomputed values.
+type DecapsulationKeyMLKEM1024 [seedSizeMLKEM]byte
+
+// GenerateKeyMLKEM1024 generates a new decapsulation key, drawing random bytes from
+// the default crypto/rand source. The decapsulation key must be kept secret.
+func GenerateKeyMLKEM1024() (DecapsulationKeyMLKEM1024, error) {
+	var dk DecapsulationKeyMLKEM1024
+	_, err := rand.Read(dk[:])
```
qmuntal: Same as before.

qmuntal (changes requested):

mlkem.go
gdams: agreed, swithed to using OpenSSL

mlkem.go
gdams: refactored, PTAL

mlkem.go
```diff
@@ -0,0 +1,361 @@
...
+
+// DecapsulationKeyMLKEM768 is the secret key used to decapsulate a shared key
+// from a ciphertext. It includes various precomputed values.
+type DecapsulationKeyMLKEM768 [seedSizeMLKEM]byte
+
+// GenerateKeyMLKEM768 generates a new decapsulation key, drawing random bytes from
+// the default crypto/rand source. The decapsulation key must be kept secret.
+func GenerateKeyMLKEM768() (DecapsulationKeyMLKEM768, error) {
+	var dk DecapsulationKeyMLKEM768
+	_, err := RandReader.Read(dk[:])
```
qmuntal: This is still not using OpenSSL MLKEM API. I was expecting something like `EVP_PKEY_Q_keygen` here.

qmuntal (changes requested):

mlkem.go
gdams: updated PTAL

qmuntal (approved):

PR #311, opened 2025-10-31
mkcgo: Add missing error symbols to {load,unload}_init_3

sergiodj:
When `initForCheckVersion` is called, and if there is an error in openssl's initialization, the codepath can reach `mkcgo_err_retrieve` which may invoke `_mkcgo_ERR_get_error_all`.  This will make the program segfault because the symbol for `ERR_get_error_all` is NULL.

We need to load a couple more symbols from openssl in order to properly handle the error and display a sane message to the user.

### DIFF
--- a/internal/ossl/zossl.c
+++ b/internal/ossl/zossl.c
@@ -629,6 +629,8 @@ void __mkcgo_unload_init_1() {
 
 void __mkcgo_load_init_3(void* handle) {
 	__mkcgo__dlsym(ERR_clear_error)
+	__mkcgo__dlsym(ERR_error_string_n)
+	__mkcgo__dlsym(ERR_get_error_all)
 	__mkcgo__dlsym(EVP_MD_fetch)
 	__mkcgo__dlsym(EVP_MD_free)
 	__mkcgo__dlsym(EVP_MD_get0_provider)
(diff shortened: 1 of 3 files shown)


internal/ossl/zossl.c
```diff
@@ -633,6 +633,8 @@ void __mkcgo_load_init_3(void* handle) {
 	__mkcgo__dlsym(EVP_MD_free)
 	__mkcgo__dlsym(EVP_MD_get0_provider)
 	__mkcgo__dlsym(EVP_default_properties_is_fips_enabled)
+	__mkcgo__dlsym(ERR_get_error_all)
```
qmuntal: `zossl.c` is autogenerated, you should update `shims.h` instead and then run `go generate ./...`.

You should update the functions definitions in this way:

```c
void ERR_error_string_n(unsigned long e, char *buf, size_t len) __attribute__((tag(""),tag("init_3"),noerror));
unsigned long ERR_get_error_all(const char **file, int *line, const char **func, const char **data, int *flags) __attribute__((tag("3"),tag("init_3"),noerror));
```

qmuntal (changes requested):

internal/ossl/zossl.c
sergiodj: Ah, gotcha.  Thanks.  Let me update the PR and force push.

sergiodj: Done.

qmuntal (approved):

qmuntal:
Thanks!

PR #330, opened 2025-12-01
Support ExpandHKDF with zero-length keys

qmuntal:
Upstream Go supports passing `0` as key length in `hkdf.Expand`, and a test exercising this case has been recently added to .

OpenSSL doesn't like zero-length inputs, so we better don't call OpenSSL in this case and directly return an empty slice (which is the correct result).

While here, remove an unused parameter.

### DIFF
--- a/hkdf.go
+++ b/hkdf.go
@@ -209,8 +209,9 @@ func ExpandHKDFOneShot(h func() hash.Hash, pseudorandomKey, info []byte, keyLeng
 		}
 		defer ossl.EVP_PKEY_CTX_free(ctx)
 		if len(out) == 0 {
-			// Nothing to do, and some engines may error on zero-length output.
-			// Call newHKDFCtx1, tought, to validate parameters.
+			// Nothing to do, so exit early.
+			// We also can't call EVP_PKEY_derive because some engines error on zero-length output.
+			// We can only exit after calling newHKDFCtx1 because we still need it to validate the parameters.
 			return out, nil
 		}
 		keylen := keyLength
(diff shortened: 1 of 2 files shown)


karianna (approved):

hkdf.go
```diff
@@ -208,6 +208,11 @@ func ExpandHKDFOneShot(h func() hash.Hash, pseudorandomKey, info []byte, keyLeng
 			return nil, err
 		}
 		defer ossl.EVP_PKEY_CTX_free(ctx)
+		if len(out) == 0 {
+			// Nothing to do, and some engines may error on zero-length output.
+			// Call newHKDFCtx1, tought, to validate parameters.
```
dagood: Typo `tought`, but I also think the reasoning can be a bit clearer:

```suggestion
			// Nothing to do, so exit early.
			// We also can't call EVP_PKEY_derive because some engines error on zero-length output.
			// We can only exit after calling newHKDFCtx1 because we still need it to validate the parameters.
```

hkdf.go
```diff
@@ -218,6 +223,11 @@ func ExpandHKDFOneShot(h func() hash.Hash, pseudorandomKey, info []byte, keyLeng
 			return nil, err
 		}
 		defer ossl.EVP_KDF_CTX_free(ctx)
+		if len(out) == 0 {
+			// Nothing to do, and some providers may error on zero-length output.
+			// Call newHKDFCtx3, tought, to validate parameters.
```
dagood:
```suggestion
			// Nothing to do, so exit early.
			// We also can't call EVP_PKEY_derive because some engines error on zero-length output.
			// We can only exit after calling newHKDFCtx3 because we still need it to validate the parameters.
```

dagood (approved):

gdams (approved):

PR #343, opened 2026-01-05
internal/fakecgo: add go generate support for updating upstream files

gdams:
Adds a maintenance mechanism to keep the fakecgo implementation in sync
with the upstream ebitengine/purego repository.


qmuntal (approved):

internal/fakecgo/update_tool.go
```diff
@@ -0,0 +1,107 @@
...
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+)
+
+const baseURL = "https://raw.githubusercontent.com/ebitengine/purego/main/internal/fakecgo"
```
dagood, line 13: I think anything that runs via `go generate` should have repeatable results, so `go generate ./...` is a safe thing to do. Basically: the commit hash should be stored somewhere. A simple UX to keep it easy to use would be to have this tool grab the latest in `main` when the locking file is not found (deleted by the user to intentionally refresh).

This also should be documented.

PR #355, opened 2026-01-22
update openssl backend to implement 386 assembly

gdams:


internal/ossl/asm_386.s
```diff
@@ -0,0 +1,134 @@
...
+
+	// Load fn pointer and call
+	MOVL	8(BP), DX		// fn from saved location
+	CALL	DX
+
+	// Restore SP to local frame
+	MOVL	BP, SP
+	JMP	ret
+
+args_3:
```
qmuntal: use the tag convention used in other arches, i.e. `_3args`. Also, the idea is that each tag should fall through the tags down them in order to avoid duplicating code.

internal/ossl/asm_386.s
```diff
@@ -0,0 +1,134 @@
+// Copyright 2026 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build !cgo
+
+#include "go_asm.h"
+#include "textflag.h"
+
+TEXT ·syscallNSystemStack_trampoline(SB),NOSPLIT,$4-0
```
qmuntal: Don't misalign the stack.

```suggestion
TEXT ·syscallNSystemStack_trampoline(SB),NOSPLIT,$16-0
```

internal/ossl/asm_386.s
```diff
@@ -0,0 +1,134 @@
...
+	CALL	DX
+
+	// Restore SP to local frame
+	MOVL	BP, SP
+	JMP	ret
+
+args_3:
+	// Save SP, push 3 args
+	MOVL	SP, BP
+	SUBL	$16, SP		// 16-byte aligned (12 bytes + padding)
```
qmuntal: Shouldn't be necessary once https://github.com/golang-fips/openssl/pull/355/changes#r2717232696 is submitted. Same for other cases.

qmuntal (changes requested):

qmuntal (approved):

PR #367, opened 2026-02-05
fakecgo: add support for s390x and loong64

gdams:
Note that this is blocked until we can get a fix for https://go.dev/issue/77449 landed


qmuntal (approved):

gdams:
I'm going to force merge even though tip CI is failing. The image will be rebuilt tonight which will contain the linker fixes for s390x

PR #374, opened 2026-02-18
Apply "go fix ./..."

dagood:
Attempting `go fix ./...` on the repo, but hit a problem. After running with and without cgo, I get an error when I try to run `GOOS=linux GOARCH=arm go fix ./...`:

```
# github.com/golang-fips/openssl/v2/internal/fakecgo
internal/fakecgo/go_libinit.go:55:20: cannot use int64(20) (constant 20 of type int64) as int32 value in range clause
```

`linux-arm` is still supported, so it seems like the loop modernizer might be missing a safeguard. I haven't checked for a golang/go issue yet. (I believe CI will catch this as a compiler error as well.)

### DIFF
--- a/chacha20poly1305_test.go
+++ b/chacha20poly1305_test.go
@@ -94,7 +94,7 @@ func TestChaCha20Poly1305Random(t *testing.T) {
 	}
 	// Some random tests to verify Open(Seal) == Plaintext
 	f := func(t *testing.T, nonceSize int) {
-		for i := range 256 {
+		for i := 0; i < 256; i++ {
 			var nonce = make([]byte, nonceSize)
 			var key [32]byte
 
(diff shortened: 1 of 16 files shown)


dagood:
* Opened https://github.com/golang/go/issues/77691

dagood:
Looks good to my eye now, but had to manually update gotip to latest golang.org/x/tools. (Generator needs to be updated to create fixed code though.)

gdams:
> Looks good to my eye now, but had to manually update gotip to latest golang.org/x/tools. (Generator needs to be updated to create fixed code though.)

@dagood I'm happy to update the generator if that would be helpful?

dagood:
Looks like the generator in this case is the code-copy from purego, which still supports Go 1.18 (which doesn't have rangeint). So, would need a special approach here to be able to apply go fix to files "owned" by this newer module while excluding the purego files (or the whole package). Given "go fix" has a known bug right now... doesn't seem worth doing yet.

The mere attempt already accomplished something useful by finding that bug. 🙂 This can wait--by definition, "go fix" doesn't make anything functionally different.

PR #377, opened 2026-02-23
CI add FreeBSD testing

gdams:
also implements the required `fakecgo` bits to support CGO-less


internal/fakecgo/go_freebsd.go
```diff
@@ -0,0 +1,100 @@
+// Copyright 2011 The Go Authors. All rights reserved.
```
qmuntal, line 1: This file is the same as `go_linux.go`. We could rename it to `go_unix.go` and add the `//go:build !cgo && !darwin` build tag, although darwin could also be retrofitted in it.

qmuntal (changes requested):

internal/fakecgo/go_freebsd.go
gdams: Agreed, this would be an upstream PR though right?

qmuntal (approved):
LGTM!

PR #397, opened 2026-03-11
Rename sliceNeverEmpty to sliceNeverNil

dagood:
`[]byte{}` is empty, but isn't nil. `sliceNeverEmpty ` -> `sliceNeverNil` seems like the most minimal rename to get this across. ([Other ideas](https://github.com/golang-fips/openssl/pull/396#discussion_r2921053643).)


qmuntal (approved):
Thanks!

PR #407, opened 2026-03-25
Add benchmark workflow

michelle-clayton-work:
Implement a workflow that runs benchmarks on different versions of go and openssl. This will run for every PR. Verified that it will catch an alloc increase: 
- https://github.com/golang-fips/openssl/pull/416

Posts results in the comments with links pointing to the specific job. Since comments have a character limit, the results will just show top regressions and shoudl show test failures.

- https://github.com/golang-fips/openssl/issues/408

### DIFF
--- a/.github/benchcheck/check.go
+++ b/.github/benchcheck/check.go
@@ -1,0 +1,295 @@
+package main
+
+import (
+	"bufio"
+	"flag"
+	"fmt"
+	"io"
+	"os"
+	"sort"
+	"strings"
+	"time"
+
+	"golang.org/x/perf/benchfmt"
+	"golang.org/x/perf/benchmath"
+)
+
+type benchKey struct {
+	Name string
+	Unit string
+}
+
+type config struct {
+	TimeThreshold float64 // minimum sec/op regression percentage
+	MinTime       float64 // minimum base time in seconds
+	Alpha         float64 // significance level
+}
+
+type regression struct {
+	Name      string
+	Unit      string
+	PctChange float64
+	PValue    float64
+	BaseVal   float64
+}
+
+func cmdCheck(args []string) {
+	fs := flag.NewFlagSet("check", flag.ExitOnError)
+	timeThreshold := fs.Float64("time-threshold", 5, "minimum sec/op regression percentage to flag")
+	minTime := fs.Duration("min-time", time.Microsecond, "minimum base time for sec/op checks")
+	alpha := fs.Float64("alpha", 0.05, "significance level for statistical tests")
+	fs.Usage = func() {
+		fmt.Fprintf(os.Stderr, "usage: benchcheck check [flags] base.txt head.txt\n\nFlags:\n")
+		fs.PrintDefaults()
+	}
+	fs.Parse(args)
+
+	if fs.NArg() != 2 {
+		fs.Usage()
+		os.Exit(2)
+	}
+
+	cfg := config{
+		TimeThreshold: *timeThreshold,
+		MinTime:       minTime.Seconds(),
+		Alpha:         *alpha,
+	}
+
+	basePath, headPath := fs.Arg(0), fs.Arg(1)
+
+	// Extract test failures from both files.
+	var failureLines []string
+	if lines, err := extractFailuresFromFile(basePath, "base: "); err == nil {
+		failureLines = append(failureLines, lines...)
+	}
+	if lines, err := extractFailuresFromFile(headPath, "head: "); err == nil {
+		failureLines = append(failureLines, lines...)
+	}
+	hasFailures := len(failureLines) > 0
+
+	// Parse benchmarks and check regressions.
+	baseValues := parseBenchmarksFromFile(basePath)
+	headValues := parseBenchmarksFromFile(headPath)
+	regressions := checkRegressions(baseValues, headValues, cfg)
+	hasRegressions := len(regressions) > 0
+
+	// Sort regressions by percentage descending.
+	sort.Slice(regressions, func(i, j int) bool {
+		return regressions[i].PctChange > regressions[j].PctChange
+	})
+
+	// Write regressions.txt.
+	var regressionLines []string
+	for _, r := range regressions {
+		if isAllocUnit(r.Unit) {
+			regressionLines = append(regressionLines, fmt.Sprintf("alloc regression: %s [%s] +%.2f%% (p=%.3f)", r.Name, r.Unit, r.PctChange, r.PValue))
+		} else {
+			regressionLines = append(regressionLines, fmt.Sprintf("time regression: %s +%.2f%% (p=%.3f, base=%.2g sec)", r.Name, r.PctChange, r.PValue, r.BaseVal))
+		}
+	}
+	writeLines("regressions.txt", regressionLines)
+	writeLines("failures.txt", failureLines)
+
+	// Write status.txt.
+	writeStatus("status.txt", hasRegressions, hasFailures)
+
+	// Print summary with GitHub Actions annotations.
+	if hasRegressions {
+		fmt.Println("::error::Benchmark regression detected — see benchstat output above for details.")
+		fmt.Println()
+		fmt.Println("=== Regressions ===")
+		for _, line := range regressionLines {
+			fmt.Println(line)
+		}
+	}
+	if hasFailures {
+		fmt.Println("::error::Test failure detected — see the 'Run benchmarks' steps for details.")
+		fmt.Println()
+		fmt.Println("=== Test failures ===")
+		for _, line := range failureLines {
+			fmt.Println(line)
+		}
+	}
+	if !hasRegressions && !hasFailures {
+		fmt.Println("No benchmark regressions or test failures detected.")
+	}
+	if hasRegressions || hasFailures {
+		os.Exit(1)
+	}
+}
+
+func extractFailuresFromFile(path, prefix string) ([]string, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+	return extractFailures(f, prefix), nil
+}
+
+// extractFailures parses go test output and returns lines related to
+// build errors, test failures, and crash traces.
+func extractFailures(r io.Reader, prefix string) []string {
+	var lines []string
+	scanner := bufio.NewScanner(r)
+	inCrash := false
+	for scanner.Scan() {
+		line := scanner.Text()
+		switch {
+		case strings.HasPrefix(line, "# "):
+			lines = append(lines, prefix+line)
+		case strings.HasPrefix(line, "--- FAIL"):
+			lines = append(lines, prefix+line)
+		case strings.HasPrefix(line, "FAIL\t"):
+			lines = append(lines, prefix+line)
+		case !inCrash && isCrashLine(line):
+			inCrash = true
+			lines = append(lines, prefix+line)
+		case inCrash && line == "":
+			inCrash = false
+			lines = append(lines, "")
+		case inCrash:
+			lines = append(lines, prefix+line)
+		}
+	}
+	return lines
+}
+
+// isCrashLine returns true for signal or panic lines (e.g. "SIGSEGV:", "panic:").
+func isCrashLine(line string) bool {
+	if strings.HasPrefix(line, "panic:") {
+		return true
+	}
+	if !strings.HasPrefix(line, "SIG") {
+		return false
+	}
+	// Match SIG followed by uppercase letters then ':'.
+	i := 3
+	for i < len(line) && line[i] >= 'A' && line[i] <= 'Z' {
+		i++
+	}
+	return i > 3 && i < len(line) && line[i] == ':'
+}
+
+func parseBenchmarksFromFile(path string) map[benchKey][]float64 {
+	f, err := os.Open(path)
+	if err != nil {
+		return nil
+	}
+	defer f.Close()
+	return parseBenchmarks(f, path)
+}
+
+func parseBenchmarks(r io.Reader, name string) map[benchKey][]float64 {
+	result := make(map[benchKey][]float64)
+	reader := benchfmt.NewReader(r, name)
+	for reader.Scan() {
+		rec := reader.Result()
+		res, ok := rec.(*benchfmt.Result)
+		if !ok {
+			continue
+		}
+		benchName := res.Name.String()
+		for _, v := range res.Values {
+			key := benchKey{Name: benchName, Unit: v.Unit}
+			result[key] = append(result[key], v.Value)
+		}
+	}
+	return result
+}
+
+func checkRegressions(base, head map[benchKey][]float64, cfg config) []regression {
+	thresholds := &benchmath.Thresholds{CompareAlpha: cfg.Alpha}
+	var regressions []regression
+
+	for key, baseVals := range base {
+		headVals, ok := head[key]
+		if !ok {
+			continue
+		}
+
+		baseSample := benchmath.NewSample(baseVals, thresholds)
+		headSample := benchmath.NewSample(headVals, thresholds)
+
+		cmp := benchmath.AssumeNothing.Compare(baseSample, headSample)
+		if cmp.P >= cmp.Alpha {
+			continue // not statistically significant
+		}
+
+		baseSummary := benchmath.AssumeNothing.Summary(baseSample, 0.95)
+		headSummary := benchmath.AssumeNothing.Summary(headSample, 0.95)
+
+		if headSummary.Center <= baseSummary.Center {
+			continue // improvement or same
+		}
+
+		if baseSummary.Center == 0 {
+			// Base is zero; only flag allocation units (sec/op can't be 0 meaningfully).
+			if isAllocUnit(key.Unit) {
+				regressions = append(regressions, regression{
+					Name:      key.Name,
+					Unit:      key.Unit,
+					PctChange: 100, // 0 → non-zero
+					PValue:    cmp.P,
+				})
+			}
+			continue
+		}
+
+		pctChange := (headSummary.Center - baseSummary.Center) / baseSummary.Center * 100
+
+		switch {
+		case isAllocUnit(key.Unit):
+			regressions = append(regressions, regression{
+				Name:      key.Name,
+				Unit:      key.Unit,
+				PctChange: pctChange,
+				PValue:    cmp.P,
+			})
+		case key.Unit == "sec/op":
+			if baseSummary.Center >= cfg.MinTime && pctChange >= cfg.TimeThreshold {
+				regressions = append(regressions, regression{
+					Name:      key.Name,
+					Unit:      key.Unit,
+					PctChange: pctChange,
+					PValue:    cmp.P,
+					BaseVal:   baseSummary.Center,
+				})
+			}
+		}
+	}
+
+	return regressions
+}
+
+func isAllocUnit(unit string) bool {
+	return unit == "B/op" || unit == "allocs/op"
+}
+
+func writeLines(path string, lines []string) {
+	if len(lines) == 0 {
+		// Write an empty file so the artifact upload doesn't skip it.
+		os.WriteFile(path, nil, 0o644)
+		return
+	}
+	f, err := os.Create(path)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "writing %s: %v\n", path, err)
+		return
+	}
+	defer f.Close()
+	for _, line := range lines {
+		fmt.Fprintln(f, line)
+	}
+}
+
+func writeStatus(path string, regression, failures bool) {
+	f, err := os.Create(path)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "writing %s: %v\n", path, err)
+		return
+	}
+	defer f.Close()
+	fmt.Fprintf(f, "regression=%v\n", regression)
+	fmt.Fprintf(f, "test_failures=%v\n", failures)
+}
--- a/.github/benchcheck/check.go
+++ b/.github/benchcheck/check.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bufio"
+	"encoding/json"
 	"flag"
 	"fmt"
 	"io"
@@ -38,8 +39,23 @@ func cmdCheck(args []string) {
 	timeThreshold := fs.Float64("time-threshold", 5, "minimum sec/op regression percentage to flag")
 	minTime := fs.Duration("min-time", time.Microsecond, "minimum base time for sec/op checks")
 	alpha := fs.Float64("alpha", 0.05, "significance level for statistical tests")
+	regressionsOut := fs.String("o-regressions", "regressions.txt", "output file for regression details")
+	failuresOut := fs.String("o-failures", "failures.txt", "output file for test failure details")
+	statusOut := fs.String("o-status", "status.json", "output file for machine-readable status (JSON)")
 	fs.Usage = func() {
-		fmt.Fprintf(os.Stderr, "usage: benchcheck check [flags] base.txt head.txt\n\nFlags:\n")
+		fmt.Fprintf(os.Stderr, `usage: benchcheck check [flags] base.txt head.txt
+
+Compare base and head benchmark results, detect regressions and test failures.
+
+Writes the following output files (paths configurable via flags):
+  regressions.txt  One-line summary per detected regression.
+  failures.txt     Extracted build errors, test failures, and crash traces.
+  status.json      Machine-readable JSON status for the report subcommand.
+
+Exits 0 if no issues are found, 1 if regressions or failures are detected.
+
+Flags:
+`)
 		fs.PrintDefaults()
 	}
 	fs.Parse(args)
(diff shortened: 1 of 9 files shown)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,373 @@
...
+              core.setOutput('pr-number', '');
+            }
+
+  bench-ubuntu:
+    needs: setup
+    strategy:
+      fail-fast: false
+      matrix:
+        go-version: [1.25.x, 1.26.x]
+        openssl-version: [3.0.1, 3.0.13, 3.1.5, 3.2.1, 3.3.0, 3.3.1, 3.4.0, 3.5.0]
```
qmuntal: Too many versions, just run one benchmark per minor version.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,373 @@
...
+
+    - name: Install benchstat
+      run: go install golang.org/x/perf/cmd/benchstat@latest
+
+    - name: Run benchmarks (base)
+      working-directory: base
+      run: |
+        export GO_OPENSSL_VERSION_OVERRIDE=${{ matrix.openssl-version }}
+        export CGO_ENABLED=1
+        go test -run='^$' -bench=. -count=7 -benchmem -timeout 30m ./... 2>&1 | tee ../base.txt || true
```
qmuntal: count=7 is useful for local benchmarks, where you done want to spend a lot of time waiting. On CI we can increase the count a bit. Let's say 10. Same applies to other bench runs.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,373 @@
...
+              });
+            }
+
+      - name: Check result
+        run: |
+          if [[ "${{ contains(needs.*.result, 'failure') }}" == "true" || "${{ contains(needs.*.result, 'cancelled') }}" == "true" ]]; then
+            echo "One or more benchmark jobs detected regressions or were cancelled."
+            exit 1
+          fi
+          echo "All benchmark jobs passed."
```
qmuntal: Add a new line here.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,373 @@
...
+          echo "found=false" >> "$GITHUB_OUTPUT"
+        fi
+
+    - name: Compare benchmarks
+      id: benchstat
+      run: |
+        benchstat base.txt head.txt | tee benchstat.txt
+        THRESHOLD="5"
+        if awk -v threshold="$THRESHOLD" '
+          /^[[:space:]]*│.*(sec\/op|B\/op|allocs\/op)/ { header=1; next }
```
qmuntal: I don't understand this regrex. But I'm ok if it works. Just one comment: it is OK to regress a bit on cpu time given that we don't have a stable machine. On the other hand, allocations shouldn't regress, not even a bit, as they don't depend on the host performance.

qmuntal (changes requested):
Looks good (with some nits)!

qmuntal (commented):
Image

Green benchmarks have an expand arrow but there is nothing to expand. Can we avoid it?

.github/workflows/benchmark.yml
michelle-clayton-work: For clarity added a small comment and the longer description is:

Here's what each line does:

/│.*(B\/op|allocs\/op)/ — Detects an allocation header row → sets header="alloc"
/│.*sec\/op/ — Detects a timing header row → sets header="time"
/^$/ — Blank line resets context (sections are separated by blanks)
match($0, /\+([0-9]+\.[0-9]+)% \(p=/, m) — Extracts the regression percentage from strings like +5.19% (p=0.001 n=10). The \+ means it only matches increases (decreases show -). m[1] captures just the number (5.19).
/[0-9].*[µm]/ — This is the unit filter, only for sec/op rows. It checks that the benchmark time contains µ (microseconds) or m (milliseconds), skipping n (nanoseconds). So 1.234µ matches but 18.31n doesn't.
The two threshold rules:

Allocations: any +X% (> 0%) → regression
CPU time: +X% where X ≥ 5, but only if the row's time is ≥ 1µs (the [µm] filter)
The [µm] filter exists because nanosecond-level benchmarks (like NewSHA256 at 18ns) show large percentage swings from normal CI noise, producing false positives.

michelle-clayton-work: Also it should flag any alloc regression with `(m[1]+0 > 0)`

michelle-clayton-work: But I will create a test branch off this and see if it correctly flags it

qmuntal (commented):
I wonder if using golang.org/x/perf to parse the bench results using Go rather than bash+awk would make this code a bit easier to maintain and to test. Could you give it a try? If it's too much of a change then we can leave it for a follow-up PR.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,379 @@
...
+        # CPU time regressions (sec/op): >= 5% fails, but only for benchmarks >= 1µs
+        # (benchstat units: n=nano, µ=micro, m=milli; /[0-9].*[µm]/ skips ns-level noise).
+        if awk '
+          /^[[:space:]]*│.*(B\/op|allocs\/op)/ { header="alloc"; next }
+          /^[[:space:]]*│.*sec\/op/ { header="time"; next }
+          /^$/ { header="" }
+          header == "alloc" && match($0, /\+([0-9]+\.[0-9]+)% \(p=/, m) { if (m[1]+0 > 0) found=1 }
+          header == "time" && /[0-9].*[µm]/ && match($0, /\+([0-9]+\.[0-9]+)% \(p=/, m) { if (m[1]+0 >= 5) found=1 }
+          END { exit !found }
+        ' benchstat.txt; then
```
dagood: A thought on this: we could move it to a tool written in Go for maintainer familiarity and perhaps a more structured approach.

*Potentially*, a Go tool could also use https://pkg.go.dev/golang.org/x/perf/benchmath etc. directly, rather than running https://github.com/golang/perf/blob/master/cmd/benchstat/main.go and parsing the intended-for-humans output.

dagood: A middle ground: there is a `-format csv` that might help simplify machine parsing. (I haven't tried it. We'd probably want to run benchstat twice in this case, once for parsing and again for the comment content.)

dagood: Oh, somehow I missed that @qmuntal already commented basically the same idea to use the package, sorry! 😄

michelle-clayton-work:
@qmuntal @dagood let me know what you think of the current change. Using x/perf is not too complicated, but it means adding it to go.mod which I wasn't sure if that was ideal considering it is not actually needed for the code but just CI so kept it in a separate go.mod

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,355 @@
...
+      working-directory: base
+      run: |
+        export CGO_ENABLED=1
+        go test -run='^$' -bench=. -count=10 -benchmem -timeout 60m ./... 2>&1 | tee ../base.txt || true
+
+    - name: Run benchmarks (head)
+      working-directory: head
+      run: |
+        export CGO_ENABLED=1
+        go test -run='^$' -bench=. -count=10 -benchmem -timeout 60m ./... 2>&1 | tee ../head.txt || true
```
dagood: Seems like a few decent chunks of duplication that could be shared somehow between bench-azl3 and bench-ubuntu?

dagood (commented):
Good approach! For the backend repos in the microsoft org, we can put the tool in microsoft/go-infra to share (as a normal cmd/ tool) without needing to split out the go.mod there.

michelle-clayton-work:
Putting all the benchstat results in a PR comment would cause it to go over the word count limit. Probably need discussions on what information we want to be in the comment / how to truncate to stay within the limits but still have all the necessary information

.github/benchcheck/main.go
```diff
@@ -0,0 +1,181 @@
+// benchcheck reads two Go benchmark result files and detects regressions.
+//
+// Usage: benchcheck [flags] base.txt head.txt
```
dagood: Seems like this should show up in the `-h` output?

I like the `mkcgo` approach of having a `func usage() {` near the top so it's still easy to find when just reading code.

qmuntal (changes requested):
This is looking great!

I'm still seeing too much bash/js in the benchmark.yml, though. Can we expand benchcheck to work better with github actions? For example, it could directly print all the output that is currently generated in the "📊 RESULTS" step. Also, it could Generate the `status.txt` that is currently generated in the "Save status" step. It could also build the final repo.

Think about benchcheck as the implementation of the whole benchmark workflow, being the bechmark.yml file just a coordinator that calls benchcheck with the appropiate parameters.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,343 @@
...
+    steps:
+    - name: Install build tools (Ubuntu)
+      if: matrix.platform == 'ubuntu'
+      run: sudo apt-get update && sudo apt-get install -y build-essential
+
+    - name: Install build tools (AZL3)
+      if: matrix.platform == 'azl3'
+      run: |
+        sed -i 's/repo_gpgcheck=1/repo_gpgcheck=0/' /etc/yum.repos.d/*.repo
+        tdnf install -y --nogpgcheck gcc binutils kernel-headers glibc-devel openssl-devel git tar ca-certificates
```
dagood: Why is gpgcheck being turned off?

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,343 @@
...
+      id: test-failures
+      run: |
+        extract_failures() {
+          local file=$1 prefix=$2
+          # Extract a compact crash/failure summary including the first goroutine's
+          # stack trace (the one that crashed). Stops at the first blank line after
+          # the crash to avoid dumping all goroutines.
+          awk '
+            # Build errors
+            /^# / { print; next }
```
dagood: There's still a few decently large chunks of awk, shell, js, and jq scripts in here. I suppose for a dev-time workflow the downsides aren't all that bad (can't test it, less maintainer familiarity, lang tooling doesn't work as well embedded in yml). Maybe the benefit of having it all in a single file is actually worthwhile. It's mainly piping and filtering, at the end of the day--I do think that `benchcheck` was the most important bit to extract. Not sure if it's worth doing more, but seemed worth commenting on.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,343 @@
...
+            # Test failures
+            /^--- FAIL/ { print; next }
+            # Package FAIL lines
+            /^FAIL\t/ { print; next }
+            # Start of a crash: capture signal, panic, and first goroutine stack
+            /^SIG[A-Z]+:/ || /^panic:/ {
```
dagood: Could use test `-json` or `test2json` to make this more structured using official tools. (Goal being that it's easier to deal with breaking changes and edge cases.)

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,343 @@
...
+    - name: Install build tools (AZL3)
+      if: matrix.platform == 'azl3'
+      run: |
+        sed -i 's/repo_gpgcheck=1/repo_gpgcheck=0/' /etc/yum.repos.d/*.repo
+        tdnf install -y --nogpgcheck gcc binutils kernel-headers glibc-devel openssl-devel git tar ca-certificates
+
+    - name: Install OpenSSL
+      if: matrix.openssl-version != ''
+      run: |
+        git clone --depth 1 --filter=blob:none --sparse ${{ github.server_url }}/${{ github.repository }} _openssl_setup
```
dagood: Is the goal here to specifically use the latest version of the `v2` branch, rather than `head` or `base`? I would think that `head` would be a good place to use the OpenSSL setup scripts from.

I think that `head` would let you e.g. perf test a new version of OpenSSL with new support added in `head`, so it might actually be better?

michelle-clayton-work: head makes sense to me. Will update!

.github/workflows/benchmark.yml
michelle-clayton-work: I think it was just related to azl3 being shipped without the gpg keys

.github/benchcheck/check.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+func cmdCheck(args []string) {
+	fs := flag.NewFlagSet("check", flag.ExitOnError)
+	timeThreshold := fs.Float64("time-threshold", 5, "minimum sec/op regression percentage to flag")
+	minTime := fs.Duration("min-time", time.Microsecond, "minimum base time for sec/op checks")
+	alpha := fs.Float64("alpha", 0.05, "significance level for statistical tests")
+	fs.Usage = func() {
+		fmt.Fprintf(os.Stderr, "usage: benchcheck check [flags] base.txt head.txt\n\nFlags:\n")
+		fs.PrintDefaults()
```
dagood, line 59: It's unlikely that someone will read this program without the full context of benchmark.yml, but I think it would be nice to explain here that it'll be writing some files to disk, and generally what the files are for.

Also (and kind of alternatively), this program could write the results to stdout by default and accept output flags to put them in files for you. `-o-regressions regressions.txt` etc. This helps clarify the benchmark.yml file because you can ctrl-f the filename in the upload step and see where they're supposed to be coming from.

.github/workflows/benchmark.yml
dagood: Interesting... would it make sense to use `mcr.microsoft.com/oss/go/microsoft/golang:1.25-azurelinux3.0` instead, like the test workflow does? Might be able to remove this step entirely?

FWIW, I don't get any gpg problems when I use this `mcr.microsoft.com/azurelinux/base/core:3.0` image locally. Still seems odd. 😄

.github/benchcheck/check.go
```diff
@@ -0,0 +1,295 @@
...
+		MinTime:       minTime.Seconds(),
+		Alpha:         *alpha,
+	}
+
+	basePath, headPath := fs.Arg(0), fs.Arg(1)
+
+	// Extract test failures from both files.
+	var failureLines []string
+	if lines, err := extractFailuresFromFile(basePath, "base: "); err == nil {
+		failureLines = append(failureLines, lines...)
```
dagood: Is dropping non-nil errors intentional? A typical API for this that integrates error handling reasonably would be:

```go
	var failureLines []string
	if failureLines, err = appendFailuresFromFile(failureLines, basePath, "base: "); err != nil {
		// handle err
```

.github/benchcheck/check.go
```diff
@@ -0,0 +1,295 @@
...
+			inCrash = true
+			lines = append(lines, prefix+line)
+		case inCrash && line == "":
+			inCrash = false
+			lines = append(lines, "")
+		case inCrash:
+			lines = append(lines, prefix+line)
+		}
+	}
+	return lines
```
dagood: Missing scanner error handling when the scan is done (or failed)?

.github/benchcheck/check.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+}
+
+func writeStatus(path string, regression, failures bool) {
+	f, err := os.Create(path)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "writing %s: %v\n", path, err)
+		return
+	}
+	defer f.Close()
```
dagood: Missing error handling. Consider `os.WriteFile` to simplify?

.github/benchcheck/report.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	fmt.Print(output)
+}
+
+type jobStatus struct {
+	regression   bool
+	testFailures bool
+	benchError   bool
+	failed       bool // derived: any of the above
```
dagood, line 107: Consider a method instead of repeating `s.failed = true` below?

.github/benchcheck/report.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+func readJobStatus(path string) jobStatus {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		// Missing status file means the job errored before writing it.
```
dagood: Couldn't the error be something other than `errors.Is(err, os.ErrNotExist)`?

.github/benchcheck/report.go
```diff
@@ -0,0 +1,155 @@
...
+		return jobStatus{failed: true, benchError: true}
+	}
+	var s jobStatus
+	for _, line := range strings.Split(string(data), "\n") {
+		switch {
+		case line == "regression=true":
```
dagood: This isn't a *lot* of code, but I feel like changing the payload to JSON would still reduce it and standardize the serialization/deserialization.

.github/workflows/benchmark.yml
```diff
@@ -235,64 +166,12 @@ jobs:
...
-                echo "</details>"
-                echo ""
-              else
-                echo ":white_check_mark: \`$label\` · [results]($job_url)"
-              fi
-            done
-          } > report.md
+          go build -C head/.github/benchcheck -o "$PWD/benchcheck" .
+          ./benchcheck report \
+            --run-url="$RUN_URL" \
```
dagood: This works, but Go flags are usually written `-run-url`.

.github/benchcheck/check.go
michelle-clayton-work: added some -o-... flags!

.github/benchcheck/check.go
michelle-clayton-work: updated

.github/benchcheck/report.go
michelle-clayton-work: Added func (s *jobStatus) setFailed(field *bool)

.github/benchcheck/report.go
michelle-clayton-work: updated so  readJobStatus checks errors.Is(err, os.ErrNotExist) specifically — other errors get logged to stderr before returning the error status

.github/workflows/benchmark.yml
michelle-clayton-work: This is the error message I get -- I'm not sure what the resolution would be here

```
Run tdnf install -y openssl-devel
Loaded plugin: tdnfrepogpgcheck
Refreshing metadata for: 'Azure Linux Official Cloud Native 3.0 x86_64'
repo md signature check: No public key
Error: TDNFVerifySignature 2004
Plugin error: repogpgcheck plugin error: failed to verify signature

Error(2004) : Unknown error 404
Error: Failed to synchronize cache for repo 'Azure Linux Official Cloud Native 3.0 x86_64'
Disabling Repo: 'Azure Linux Official Cloud Native 3.0 x86_64'
Refreshing metadata for: 'Azure Linux Official Base 3.0 x86_64'
repo md signature check: No public key
Error: TDNFVerifySignature 2004
Plugin error: repogpgcheck plugin error: failed to verify signature

Error(2004) : Unknown error 404
Error: Failed to synchronize cache for repo 'Azure Linux Official Base 3.0 x86_64'
Disabling Repo: 'Azure Linux Official Base 3.0 x86_64'
Refreshing metadata for: 'Azure Linux Official Microsoft Open-Source 3.0 x86_64'
repo md signature check: No public key
Error: TDNFVerifySignature 2004
Plugin error: repogpgcheck plugin error: failed to verify signature

Error(2004) : Unknown error 404
Disabling Repo: 'Azure Linux Official Microsoft Open-Source 3.0 x86_64'
Refreshing metadata for: 'Azure Linux Official Microsoft Non-Open-Source 3.0 x86_64'
Error: Failed to synchronize cache for repo 'Azure Linux Official Microsoft Open-Source 3.0 x86_64'
repo md signature check: No public key
Error: TDNFVerifySignature 2004
Plugin error: repogpgcheck plugin error: failed to verify signature

Error(2004) : Unknown error 404
Error: Failed to synchronize cache for repo 'Azure Linux Official Microsoft Non-Open-Source 3.0 x86_64'
Disabling Repo: 'Azure Linux Official Microsoft Non-Open-Source 3.0 x86_64'
openssl-devel package not found or not installed
Error(1011) : No matching packages
```

dagood: Ah, it's because GitHub Actions changes HOME, but the image isn't expecting it. See https://github.com/Azure/azure-cli/issues/29835. Repros locally if I include `-e "HOME=/github/home"`.

This worked for me: `export GNUPGHOME=/root/.gnupg`

dagood: Tracking issue on AzL's side: https://github.com/microsoft/azurelinux/issues/2812

qmuntal (approved):
This seems good to go! We can always improve things in future PRs as we find improvement opportunities.

.github/benchcheck/check.go
```diff
@@ -0,0 +1,314 @@
...
+func isAllocUnit(unit string) bool {
+	return unit == "B/op" || unit == "allocs/op"
+}
+
+func writeLines(path string, lines []string) {
+	data := []byte(strings.Join(lines, "\n") + "\n")
+	if len(lines) == 0 {
+		data = nil
+	}
+	if err := os.WriteFile(path, data, 0o644); err != nil {
```
dagood, line 292: Consider adding a quick `os.MkdirAll(filepath.Dir(path))` before this so we don't at some point(s) in the future end up wasting a dev iteration on having to explicitly pre-create a directory in whatever script is wrapping this tool.

.github/benchcheck/check.go
```diff
@@ -0,0 +1,314 @@
...
+		fmt.Fprintf(os.Stderr, "writing %s: %v\n", path, err)
```
dagood, line 293: It strikes me as odd to only write out to stderr. I think funcs like this should almost always return the error (with additional context added in via fmt.Errorf) so the caller can decide what to do.

About handling overall, I think that a failure to write these files should end up causing the command to exit 1 rather than 0, right? These files aren't just supplemental, so I don't see a reason the command should behave as if it was a success.

(It is reasonable to let the command try to finish up what it can, though, and use `errors.Join` to accumulate.)

.github/benchcheck/report.go
dagood: What came to mind for me was a getter `func (s *jobStatus) failed() { return s.regression || s.testFailures || s.benchError }`, this is nifty though. 🙂

.github/benchcheck/check.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+}
+
+// Status is the machine-readable status written by check and read by report.
+type Status struct {
+	Regression     bool `json:"regression"`
+	TestFailures   bool `json:"test_failures"`
+	BenchmarkError bool `json:"benchmark_error"`
+}
```
dagood, line 302: (Near-)Duplicate of `type jobStatus` defined in report.go?

.github/benchcheck/report.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	if status.BenchmarkError {
+		s.setFailed(&s.benchError)
+	}
+	return s
+}
+
+func loadJobURLs(path string) map[string]string {
+	if path == "" {
+		return nil
```
dagood, line 144: Does the caller really expect nil (not even an empty map) to be a possible output?

.github/benchcheck/report.go
```diff
@@ -0,0 +1,166 @@
...
+			urls[parts[0]] = parts[1]
+		}
+	}
+	return urls
+}
+
+func readFileContent(path string) string {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return ""
```
dagood, line 163: It doesn't seem safe to ignore all possible errors here. (Similar in `loadJobURLs`.)

.github/benchcheck/check.go
```diff
@@ -0,0 +1,314 @@
...
+
+func cmdCheck(args []string) {
+	fs := flag.NewFlagSet("check", flag.ExitOnError)
+	timeThreshold := fs.Float64("time-threshold", 5, "minimum sec/op regression percentage to flag")
+	minTime := fs.Duration("min-time", time.Microsecond, "minimum base time for sec/op checks")
+	alpha := fs.Float64("alpha", 0.05, "significance level for statistical tests")
+	regressionsOut := fs.String("o-regressions", "regressions.txt", "output file for regression details")
+	failuresOut := fs.String("o-failures", "failures.txt", "output file for test failure details")
+	statusOut := fs.String("o-status", "status.json", "output file for machine-readable status (JSON)")
```
dagood, line 44: I think these should default to `""` and thus not write files by default. This means the benchmark.yml is forced to specify them by name, making benchmark.yml clearer.

Right now, reading benchmark.yml, it appears that e.g. `status.json` is written to by `echo '{"regression":false,"test_failures":false,"benchmark_error":true}' > status.json` and stays that way until `Upload results` when it's uploaded.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,218 @@
...
+      if: matrix.openssl-version != ''
+      run: sudo sh head/scripts/openssl.sh ${{ matrix.openssl-version }}
+
+    - uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
+      with:
+        go-version: ${{ matrix.go-version }}
+        cache-dependency-path: head/go.mod
+
+    - name: Install benchstat
+      run: go install golang.org/x/perf/cmd/benchstat@latest
```
dagood, line 96: Consider putting this dependency (as a `tool`) in the benchcheck go.mod so `latest` can't float to a version that could somehow take advantage of `pull-requests: write`.

dagood (commented):
I agree this is good to have--some suggestions for polishing it a bit.

.github/workflows/benchmark.yml
michelle-clayton-work: Sounds good! Will put this in a followup issue! Maybe I can fix it for the backends implementation

PR #415, opened 2026-04-07
ci: add 1x benchmark runs to ensure they pass

dagood:
Add 1x (one iteration) benchmark runs to each platform to make sure they pass everywhere. This helps make sure that PRs that don't intend to change the benchmarks don't break them, making it harder to get a proper benchmark run down the line.

This could also be considered a simple test case.

It's probably not necessary to do this on every platform, because the benchmarks themselves shouldn't be the only place where platform-specific behavior causes a test failure. But perhaps it's worth verifying anyway as long as it doesn't get in the way?


qmuntal (approved):
LGTM!

dagood:
Hits some failures... will try ~~adding skips to get a baseline.~~(Actually, found the root cause.)

qmuntal (approved):

PR #417, opened 2026-04-08
Stop passing nil pointer to OSSL_PARAM_BLD_push_octet_string

chouquette:
This should highlight a breaking API change in OpenSSL: https://github.com/openssl/openssl/commit/e4610b43539b

### DIFF
--- a/params.go
+++ b/params.go
@@ -108,6 +108,10 @@ func (b *paramBuilder) addUTF8String(name cString, value *byte, size int) {
 	}
 }
 
+// emptyOctetStringBuf serves as a non-nil sentinel pointer for
+// zero-length octet strings passed to OpenSSL.
+var emptyOctetStringBuf byte
+
 // addOctetString adds an octet string to the builder.
 // The value is pinned and will be unpinned when the builder is freed.
 func (b *paramBuilder) addOctetString(name cString, value []byte) {
@@ -116,6 +120,10 @@ func (b *paramBuilder) addOctetString(name cString, value []byte) {
 	}
 	if len(value) != 0 {
 		b.pinner.Pin(&value[0])
+	} else {
+		// OpenSSL >= 3.5.6 rejects a NULL buf in OSSL_PARAM_BLD_push_octet_string
+		// even when bsize is 0 (see crypto/param_build.c).
+		value = unsafe.Slice(&emptyOctetStringBuf, 0)
 	}
 	if _, err := ossl.OSSL_PARAM_BLD_push_octet_string(b.bld, name.ptr(), value); err != nil {
 		b.err = addParamError{name.str(), err}
--- a/params.go
+++ b/params.go
@@ -108,10 +108,6 @@ func (b *paramBuilder) addUTF8String(name cString, value *byte, size int) {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


qmuntal (changes requested):
New versions should also be added to https://github.com/golang-fips/openssl/blob/v2/scripts/openssl.sh.

chouquette:
Indeed, I'll make the change.

Full disclosure, I was hoping to run a workflow with the new version before commiting the associated fix

qmuntal (commented):
What if we omit calling `OSSL_PARAM_BLD_push_octet_string` altogether when the buffer length is zero? Would that work?

chouquette:
> What if we omit calling OSSL_PARAM_BLD_push_octet_string altogether when the buffer length is zero? Would that work?

That was my initial intent, but AFAIU some parameters (`_OSSL_KDF_PARAM_SALT` for instance) need to differentiate between "explicitly provided yet empty" and "not provided".

qmuntal:
> > What if we omit calling OSSL_PARAM_BLD_push_octet_string altogether when the buffer length is zero? Would that work?
> 
> That was my initial intent, but AFAIU some parameters (`_OSSL_KDF_PARAM_SALT` for instance) need to differentiate between "explicitly provided yet empty" and "not provided".

Then it would be better to provide an empty slice in such cases. no? `unsafe.SliceData` will return a non-nil pointer in such case, but the length will still be zero.

params.go
```diff
@@ -116,6 +120,10 @@ func (b *paramBuilder) addOctetString(name cString, value []byte) {
 	}
 	if len(value) != 0 {
 		b.pinner.Pin(&value[0])
+	} else {
+		// OpenSSL >= 3.5.6 rejects a NULL buf in OSSL_PARAM_BLD_push_octet_string
+		// even when bsize is 0 (see crypto/param_build.c).
+		value = unsafe.Slice(&emptyOctetStringBuf, 0)
 	}
 	if _, err := ossl.OSSL_PARAM_BLD_push_octet_string(b.bld, name.ptr(), value); err != nil {
```
xnox, line 124: If the value provided is empty; why is this being called at all?
It feels like everything should still work without pushing such build parameter if there is nothing to provide.
unless it does not, in which case we have more bugs.

xnox: also possibly this should be fixed in openssl upstream as well.

chouquette:
> Then it would be better to provide an empty slice in such cases. no? unsafe.SliceData will return a non-nil pointer in such case, but the length will still be zero.

Fair enough, I just pushed a commit with that change (let me know if I should squash it in the commit introducing the change)

params.go
chouquette: Fair question, that's also what I was hoping for, but as mentioned above some options bear a different meaning if set to an empty value or not set at all

chouquette: > also possibly this should be fixed in openssl upstream as well.

Also fair, since this was explicitly introduced in the most recent releases I assumed it to be a thought out change, but at the very least it might be worth pointing out that it breaks existing code
I'll open an issue to start the discussion there

qmuntal (approved):

PR #418, opened 2026-04-09
Short-circuit nil octet string params

qmuntal:
This is another approach to fix #417 in which we keep the original caller intent: empty slices are passed to OpenSSL, nil slices are just skipped.

### DIFF
--- a/params.go
+++ b/params.go
@@ -115,6 +115,11 @@ func (b *paramBuilder) addOctetString(name cString, value []byte) {
 		return
 	}
 	if value == nil {
+		// Short-circuit a nil slice: don't pass anything at all to OpenSSL.
+		// OpenSSL 3.5.6 raises an error when passed null, and expects users
+		// to not call this function at all in this case.
+		// See https://github.com/openssl/openssl/issues/30728
+		//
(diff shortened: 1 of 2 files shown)


params.go
```diff
@@ -114,12 +114,13 @@ func (b *paramBuilder) addOctetString(name cString, value []byte) {
 	if !b.check() {
 		return
 	}
+	if value == nil {
+		// Don't short-circuit empty slices, as they might have a meaning.
+		// For example, in KDFs an empty salt is different from a nil salt.
```
dagood, line 124: I think starting with a positive explanation rather than what it doesn't do makes it easier to fully understand on first read:

```suggestion
		// Short-circuit a nil slice: don't pass anything at all to OpenSSL.
		// OpenSSL 3.5.6 raises an error when passed null, and expects users
		// to not call this function at all in this case.
		// See https://github.com/openssl/openssl/issues/30728
		//
		// Don't short-circuit empty slices, as they might have a meaning.
		// For example, in KDFs an empty salt is different from a nil salt.
```

dagood (approved):

PR #421, opened 2026-04-14
Support OpenSSL 4

michelle-clayton-work:
Add support for OpenSSL 4

- https://github.com/microsoft/go/issues/2187

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         go-version: [1.25.x, 1.26.x]
-        openssl-version: [1.1.1, 3.0.1, 3.0.13, 3.1.5, 3.2.1, 3.3.0, 3.3.1, 3.4.0, 3.5.0, 3.5.6]
+        openssl-version: [1.1.1, 3.0.1, 3.0.13, 3.1.5, 3.2.1, 3.3.0, 3.3.1, 3.4.0, 3.5.0, 3.5.6, 4.0.0-alpha1]
         host: [ubuntu-22.04, ubuntu-24.04-arm]
     runs-on: ${{ matrix.host }}
     steps:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         go-version: [1.25.x, 1.26.x]
-        openssl-version: [1.1.1, 3.0.1, 3.0.13, 3.1.5, 3.2.1, 3.3.0, 3.3.1, 3.4.0, 3.5.0, 3.5.6, 4.0.0-alpha1]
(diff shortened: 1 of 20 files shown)


scripts/openssl.sh
```diff
@@ -86,6 +86,13 @@ case "$version" in
         make="build_libs"
         install="install_fips"
         ;;
+    "4.0.0")
+        tag="openssl-4.0.0-alpha1";
+        sha256="fab77ba96a03f7d4c98d6c885396b1ff0a06cae0070bdcd84977f7deb212bc11"
+        config="enable-fips"
+        make="build_libs"
+        install="install_fips"
+        ;;
```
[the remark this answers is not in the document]
michelle-clayton-work: updated -- we will have to change this when 4.0.0 final version is released

.github/workflows/test.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         go-version: [1.25.x, 1.26.x]
-        openssl-version: [1.1.1, 3.0.1, 3.0.13, 3.1.5, 3.2.1, 3.3.0, 3.3.1, 3.4.0, 3.5.0, 3.5.6]
+        openssl-version: [1.1.1, 3.0.1, 3.0.13, 3.1.5, 3.2.1, 3.3.0, 3.3.1, 3.4.0, 3.5.0, 3.5.6, 4.0.0-alpha1]
```
qmuntal: OpenSSL 4.0.0 has just been released: https://github.com/openssl/openssl/releases/tag/openssl-4.0.0. No need to use alpha1 anymore.

openssl.go
```diff
@@ -73,10 +73,17 @@ func errUnsupportedVersion() error {
 }
 
 // checkMajorVersion panics if the current major version is not expected.
+// For expected >= 3, any major version >= expected is accepted, since OpenSSL
+// maintains API and ABI compatibility within a major version series,
+// and newer major versions (e.g. 4.x) retain compatibility with 3.x APIs.
```
qmuntal: This is not true. OpenSSL 4 has some API and ABI breaking changes. AFAIK, they don't affect us, but I would prefer to keep this function strict and explicit. You can change the function signature to accept a variadic int slice, and then callers can pass all the expected versions.

qmuntal (changes requested):
General note: there are some `if vMajor >= 3` new occurrences, I would prefer to use `switch vMajor { case 3,4 ...` for consistency. I ack that we should improve the way we support new major versions, as in most cases being so strict is not really necessary, but let's do that in a follow-up PR.

qmuntal (changes requested):
The `.github/benchcheck/benchcheck` binary shouldn't have been commited. Add a .gitignore rule that explicitly ignores it so it doesn't happen again.

qmuntal (approved):

PR #432, opened 2026-04-17
Skip benchmarks on .github-only changes; raise time threshold to 50%

michelle-clayton-work:
Improve benchmarks to only run on relevant changes and raise threshold to avoid flaky failures 

- https://github.com/microsoft/go-lab/issues/342

### DIFF
--- a/.github/benchcheck/check.go
+++ b/.github/benchcheck/check.go
@@ -36,7 +36,7 @@ type regression struct {
 
 func cmdCheck(args []string) {
 	fs := flag.NewFlagSet("check", flag.ExitOnError)
-	timeThreshold := fs.Float64("time-threshold", 5, "minimum sec/op regression percentage to flag")
+	timeThreshold := fs.Float64("time-threshold", 50, "minimum sec/op regression percentage to flag")
 	minTime := fs.Duration("min-time", time.Microsecond, "minimum base time for sec/op checks")
 	alpha := fs.Float64("alpha", 0.05, "significance level for statistical tests")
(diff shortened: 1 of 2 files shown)


dagood (approved):
It could be interesting to grab data from multiple runs on the same (or effectively the same) commits over time to capture some variance per benchmark. Could apply more specific stats techniques then, or just some more fine tuning.

Other ideas:
* Automatically rerun a few times if it fails, perhaps in a way that will reserve a fresh Actions agent in case we got a dud. (Although `v2` could have gotten an unusually fast agent... some complications here depending on how well we want to do.)
* When comparing results, pick the least-performant base sample and best-performing head sample, or something like that, rather than comparing `Center`s.
* Actually... does the current data have info about variance? There might be more to we can do there without changing anything about how the data is collected.

michelle-clayton-work:
> It could be interesting to grab data from multiple runs on the same (or effectively the same) commits over time to capture some variance per benchmark. Could apply more specific stats techniques then, or just some more fine tuning.
> 
> Other ideas:
> 
> * Automatically rerun a few times if it fails, perhaps in a way that will reserve a fresh Actions agent in case we got a dud. (Although `v2` could have gotten an unusually fast agent... some complications here depending on how well we want to do.)
> * When comparing results, pick the least-performant base sample and best-performing head sample, or something like that, rather than comparing `Center`s.
> * Actually... does the current data have info about variance? There might be more to we can do there without changing anything about how the data is collected.

Will add these to the improvement issue! Once these benchmarks are in a stable place I hope to integrate continuous benchmarking which shows us a timechart of benchmarks so we can see how they vary over an extended period of time!

PR #436, opened 2026-04-23
fips without enforcement support for hkdf and hmac

xnox:
Always set key-check:0 hmac and kdf param.

This enables allowed and non-approved usage of short keys, or IKM as required for TLS and legacy/deprecated calculations.

Strict enforcement of these is done at the higher-level in fips-only mode at the go standard library level, meaning by the time one uses the openssl bindings, one really needs these calculations to be successful.

With this changes, existing test-suite, passes correctly with more strict/recent openssl fips provider builds configured in pedantic mode.


qmuntal (changes requested):
Thanks. Please put the `fips140.Enforced()` call in a file gated behind a `go1.26` build tag. Priors versions should return true, to keep behavior the same.

xnox:
note still draft, as i am working through bubbling this up in the toolchain and end application to see if it will all work as intended.

xnox:
> ':white_check_mark: **No significant regressions detected**\n' +

Is the benchmark conclusion; but it failed to post a comment.

qmuntal:
I would prefer not to add yet another optional "mode", we already have too many moving pieces (openssl versions, providers, distors patches, ...). I do agree that this PR improves compatibility with the Go standard library, though.

Here is another approach.

The Go crypto packages already error/panic when `fips140=fipsonly` is set, and that enforcement happens before calling into the FIPS module, that is, at the same layer we put our patches. With this in mind, we could always   call OpenSSL APIs as if we were not enforcing FIPS (which is the expected behavior when using Go crypto), and rely on the Go crypto packages (may require some additional patching) to bail out if FIPS is enforced and a non-compliant parameter is used.

FIPS enforcement is not part of the FIPS certification, so this approach wouldn't break compliance. Also, it will bring some consistency across backends and the standard library itself.

So, this means that this PR is almost good, but it should pass `OSSL_KDF_PARAM_FIPS_KEY_CHECK` unconditionally, + documenting why we do that every time we do it.

Finally, this change targets Go 1.27, not Go 1.26.

xnox:
> I would prefer not to add yet another optional "mode", we already have too many moving pieces (openssl versions, providers, distors patches, ...). I do agree that this PR improves compatibility with the Go standard library, though.
> 
> Here is another approach.
> 
> The Go crypto packages already error/panic when `fips140=fipsonly` is set, and that enforcement happens before calling into the FIPS module, that is, at the same layer we put our patches. With this in mind, we could always call OpenSSL APIs as if we were not enforcing FIPS (which is the expected behavior when using Go crypto), and rely on the Go crypto packages (may require some additional patching) to bail out if FIPS is enforced and a non-compliant parameter is used.
> 
> FIPS enforcement is not part of the FIPS certification, so this approach wouldn't break compliance. Also, it will bring some consistency across backends and the standard library itself.

I agree.

There will be some trivial differences, which are inconsequential - for example password lengths are different between different CMVP modules (most pick 8-characters, but some pick 14-characters). Hence in only mode, if we always pass the key-check verification relaxation the module will raise dynamic indicator of non-approved usage.

> 
> So, this means that this PR is almost good, but it should pass `OSSL_KDF_PARAM_FIPS_KEY_CHECK` unconditionally, + documenting why we do that every time we do it.
> 
> Finally, this change targets Go 1.27, not Go 1.26.

Ack, we can release this update with 1.27 based toolchains.

Note, if the parameter is unconditionally passed, this code change becomes go version agnostic and distributors can backport this change to any prior versions as needed.

xnox:
@qmuntal updated!

qmuntal (approved):
Thanks!

PR #439, opened 2026-04-30
Improve versioncheck analyzer

michelle-clayton-work:
This pull request introduces a static analyzer to enforce justification for OpenSSL version checks and refines version support logic in the backend. The main changes include adding the `checkversioncheck` analyzer (with tests and fixtures), updating the workflow to run this check, and improving how untested OpenSSL major versions are handled via a GODEBUG flag. The OpenSSL backend logic is also updated to generalize support for 3+ majors and to centralize the list of tested majors.

**New static analyzer to enforce version check justification:**

* Added `cmd/checkversioncheck` tool, which scans backend code for calls to version-checking helpers (like `minor()`, `patch()`, `versionAtOrAbove()`, `checkMajorVersion()`) inside OpenSSL 3+ branches, and requires a `//openssl:versioncheck <reason>` marker comment immediately above each call. The analyzer exits non-zero if unjustified checks are found.
* Added comprehensive unit tests for the analyzer in `checkversioncheck_test.go`, with positive and negative fixtures in `testdata/okversion` and `testdata/badversion`. [[1]](diffhunk://#diff-c270e24397c43e2be7272a8c85f98570e99312c3b07a70d1e354dca9a52b2ab2R1-R141) [[2]](diffhunk://#diff-1e3f69e0f660df7b28f9458c2d4db02d19bd29bac5bdc9afcea8d8488f047100R1-R70) [[3]](diffhunk://#diff-c1cd5777329acbce5b811fe68e2063b1703997987c243a88a77e8a965a52efaaR1-R56)

**CI integration:**

* Updated `.github/workflows/test.yml` to run the new analyzer as part of CI, ensuring the justification convention is enforced on every PR.

**OpenSSL version support logic improvements:**

* Centralized the list of tested OpenSSL major versions in `osslsetup.go` as `testedMajors`, and added logic to allow untested 3+ majors only if `GODEBUG=ms_opensslallowuntested=1` is set. The GODEBUG parsing matches Go's internal behavior.
* Updated `openLibrary` and `initForCheckVersion` to generalize handling of 3+ majors, using the new `testedMajors` list and allowing future 3+ majors with the GODEBUG override. [[1]](diffhunk://#diff-ab81d66c2c52cdd2e5e5f06d570f7523329be4b97084290e575e6f4bea3615edL148-R163) [[2]](diffhunk://#diff-ab81d66c2c52cdd2e5e5f06d570f7523329be4b97084290e575e6f4bea3615edL75-R84)
* Added tests for the GODEBUG parsing and for the sortedness of `testedMajors` in `osslsetup_test.go`.

These changes ensure OpenSSL version checks are consistently justified and documented, and that backend support for new OpenSSL major versions is both robust and easily auditable.

- https://github.com/microsoft/go/issues/2253


.github/workflows/test.yml
```diff
@@ -10,6 +10,21 @@ concurrency:
 
 name: Test
 jobs:
+  checkversioncheck:
+    # Lint pass: enforce the //openssl:versioncheck convention from
+    # microsoft/go ADR 0018. Runs once (no OpenSSL matrix), since it's
```
qmuntal: Don't mention internal ADR's in this repo.

osslsetup/osslsetup.go
```diff
@@ -4,7 +4,9 @@ package osslsetup
 
 import (
 	"errors"
+	"os"
```
qmuntal: Don't import `os` here, it will fail upstream `TestDependencies`. Use https://pkg.go.dev/syscall#Getenv instead.

qmuntal (changes requested):

cmd/checkversioncheck/main.go
```diff
@@ -0,0 +1,286 @@
+// checkversioncheck is a small static analyzer that enforces the
```
dagood: The name is fun, but it does throw me off more than I would expect... I guess this is named similarly to `errcheck` and `staticcheck`, but a tool that looks for "check version"s isn't as meaningful to me as one that looks for "version check"s, so `versioncheckcheck` would honestly be more intuitive. 😄

A tighter term is probably advisable, though: `versioncondcheck`, `versiongatecheck`, `versionguardcheck`, etc.

(It's also odd that `//openssl:versioncheck` omits the first `check`. Maybe just the tool name from an old iteration? When I read this, I expect the [idioms of a directive comment](https://go.dev/doc/comment#directives).)

michelle-clayton-work: renamed to versionguardcheck

cmd/versionguardcheck/main.go
```diff
@@ -0,0 +1,286 @@
...
+	"go/parser"
+	"go/token"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+)
+
+const markerPrefix = "//openssl:versionguard"
```
dagood: [Conventionally](https://go.dev/doc/comment#directives), I'd expect this to be `//versionguardcheck:ignore`. But when considering this, I also wondered if there's a way we should really be plugging into an existing linter, and bumped into https://pkg.go.dev/golang.org/x/tools/go/analysis. It looks like it has a subpackage that provides testing helpers, too. (There's still no common syntax to ignore vet problems, but a more standard-looking `//lint:ignore versionguardcheck <Reason>` might be more reasonable in this context?)

michelle-clayton-work: Updating the name and made an issue for the xtools part! 
- https://github.com/microsoft/go/issues/2259

michelle-clayton-work: @qmuntal did you have any extra thoughts on this?

qmuntal: Switching this tool to a Go analyzer in a future PR will be a good exercise. I'm OK with that if it doesn't add a lot of additional complexity (which I doubt it will).

qmuntal (approved):

PR #450, opened 2026-05-19
Implement PBKDF2 using EVP_KDF_derive in OpenSSL 3

qmuntal:
[PKCS5_PBKDF2_HMAC](https://docs.openssl.org/3.0/man3/PKCS5_PBKDF2_HMAC/) has strict buffer lengths limitations and has a stricter compliance checks than `crypto/pbkdf2`.

We can sidestep these limitations by using the OpenSSL 3 API `EVP_KDF_derive`, which we already use elsewhere.

@xnox

### DIFF
--- a/const.go
+++ b/const.go
@@ -54,13 +54,16 @@ const ( //checkheader:ignore
 	_OSSL_MAC_NAME_HMAC      cString = "HMAC\x00"
 
 	// KDF parameters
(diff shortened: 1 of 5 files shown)


gdams (approved):

qmuntal:
Small regression, around 0.5%, in allocated size benchmarks. That's fine.

xnox:
YES!

